PR #7, opened 2018-09-04
Add transform to dome light

radon199:
Hello,

Dome lights currently do not respect the authored transform of the light in the USD stage.

This PR gets the transform from the sceneDelegate and adds it to the rprEnvironmentLight, along with defining a new token for the HDRprDomeLight.

Tested locally against USD 18.09.

Thank you,

-Colin

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This plugin allows fast GPU or CPU accelerated viewport rendering on all OpenCL 
 
 For more details on USD, please visit the web site [here](http://openusd.org).
 
-![Alt Text](https://www.dropbox.com/s/adqexgc78v35417/ezgif-2-5eb2f04c90.gif)
+![](https://www.dropbox.com/s/adqexgc78v35417/ezgif-2-5eb2f04c90.gif?raw=1)
 
 Prerequisites
 -----------------------------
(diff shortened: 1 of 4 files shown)


sayakbiswas:
Hi @radon199 I gave the changes a quick test, but the transforms on the dome lights still don't seem to work properly. Can you share the scene you used for testing? Also, what version of RPR SDK are you on?

radon199:
Hey,

Sorry you are having trouble.

This took some digging and I think I have discovered a bug/unfinished feature in USD itself in the process.

The short story is that all lights in USD, when imaged, will only respect their local transform. This is true for lights imaged with hdSt or hdRpr.

I took part of the code from the hdSt light.cpp on the Pixar repo, there is also a similar function in the lightBase.cpp file of hdRpr

`    // Transform
    if (bits & DirtyTransform) {
        VtValue transform = sceneDelegate->Get(id, HdLightTokens->transform);
        if (transform.IsHolding<GfMatrix4d>()) {
            _params[HdLightTokens->transform] = transform;
        } else {
            _params[HdLightTokens->transform] = GfMatrix4d(1);
        }
    }`

There is a GetTransform() function in the sceneDelegate but that only works on rprims, and lights are sprims.

Therefore we need to use the Get() function and ask for the transform. This has the side effect of only getting the local transform and not the global transform down to the root.

I have included a file below that demonstrates this. In it the sphere and the light on the grid should be in the same place as they have the same transforms, but they are not, the light only respects it's own local transform relative to the world.

I will file a ticket with Pixar.

[lightTransformTest.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/2359360/lightTransformTest.zip)

bsavery:
Seems to work fine with 18.11 version of USD.  Thanks!

PR #18, opened 2019-07-02
Fix: Incorrect preferences file path generation

hshakula:
Fix incorrect preferences file path generation

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -103,11 +103,11 @@ std::string GetRprTmpDir()
 	if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, 0, appDataPath)))
 	{
 		char buff[1024] = {};
-		sprintf(buff, "%s/UsdRpr/", appDataPath);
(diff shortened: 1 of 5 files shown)


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -94,27 +94,28 @@ inline bool rprIsErrorCheck(const TfCallContext &context, const rpr_status statu
...
-
 	char appDataPath[MAX_PATH];
 	// Get path for each computer, non-user specific and non-roaming data.
 	if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, 0, appDataPath)))
 	{
-		char buff[1024] = {};
-		sprintf(buff, "%s/UsdRpr/", appDataPath);
-		return std::string(buff);
+		static char path[MAX_PATH];
+		snprintf(path, sizeof(path), "%s\\UsdRpr\\", appDataPath);
```
bsavery: Should this folder be called hdRPR or UsdRpr?  I'm slightly leaning towards hdRPR

bsavery: But please give your thoughts.

hshakula: hdRPR is closer to reality because this is actually Hydra plugin

PR #19, opened 2019-07-04
Fix: Broken render device switching

hshakula:
GUI render device switching button actually does not really change device type but simply update HdRprPreferences. To really change render device we should somehow reset current context and refill scene. Unfortunately, USD API does not provide the possibility to do some sort of hot reset of renderer plugin, like usdViewApi.RestartPlugin()

I propose shortest in time workaround for this problem but obviously not the most optimal one.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/python/rpr.py
+++ b/pxr/imaging/plugin/hdRpr/python/rpr.py
@@ -28,12 +28,13 @@ def setFilter(filter):
 	   lib.SetRprGlobalFilter(filter)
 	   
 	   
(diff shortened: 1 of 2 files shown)


bsavery:
Thanks.  I posted a question also on:

https://groups.google.com/forum/?utm_source=digest&utm_medium=email#!topic/usd-interest/SGlQyk90Mlo

bsavery:
According to Pol this is a good way to do this for now.  So We'll merge until USD api gives a better method.

bsavery:
@hshakula Can you fix the conflict with the pref directory?  Thanks

Issue #24, opened 2019-07-05
Incorrect texture coordinates for meshes from Apple's AR Quick Look gallery


hshakula:
# [Teapot](https://developer.apple.com/augmented-reality/quick-look/models/teapot/teapot.usdz)
## Embree
![Embree_teapot](https://user-images.githubusercontent.com/22181845/60716523-4cf2ce80-9f28-11e9-95c5-527331dec025.png)
## RPR
![RPR_teapot](https://user-images.githubusercontent.com/22181845/60716525-4d8b6500-9f28-11e9-8693-90102a493dc6.png)

# [Gramophone](https://developer.apple.com/augmented-reality/quick-look/models/gramophone/gramophone.usdz)
## Embree
![Embree_gramophone](https://user-images.githubusercontent.com/22181845/60716522-4cf2ce80-9f28-11e9-9902-870b1229588b.png)
## RPR
![RPR_gramophone](https://user-images.githubusercontent.com/22181845/60716524-4d8b6500-9f28-11e9-9cb0-44e9a64bddf8.png)

PR #28, opened 2019-07-10
Fix: Incorrect primvar indexing #24

hshakula:
Fix for #24

### DIFF
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -90,33 +90,94 @@ void HdRprMesh::Sync(
 	{
 		HdMeshTopology meshTopology = GetMeshTopology(sceneDelegate);
 
+		const VtIntArray & indexes = meshTopology.GetFaceVertexIndices();
+		const VtIntArray & vertexPerFace = meshTopology.GetFaceVertexCounts();
+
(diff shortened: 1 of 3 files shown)


pxr/imaging/plugin/hdRpr/mesh.cpp
```diff
@@ -90,33 +90,94 @@ void HdRprMesh::Sync(
...
+
 		VtValue value;
 		value = sceneDelegate->Get(id, HdTokens->points);
 		VtVec3fArray points = value.Get<VtVec3fArray>();
 
 		VtVec2fArray st;
-
-
-		value = sceneDelegate->Get(id, TfToken("st"));
+		float timeSamples[1] = {0.0f};
```
bsavery: What happens if ST is varying over time, does this always get the first timesample?  Or just the one at current time?

hshakula: This gets first or default timesample. Unfortunately, there is no way to get the current time from HdSceneDelegate, only UsdImagingDelegate knows it.

Let's see what USD team thinks about adding a virtual method GetTime to HdSceneDelegate. [Issue](https://github.com/PixarAnimationStudios/USD/issues/897)

hshakula: I totally missed that SamplePrimvar accepts not absolute time but relative to the current time.
So present changes do fully handle varying over time ST

PR #34, opened 2019-07-18
Support for per-face-vertex UVs, missing UVs, and explicit normals

elrond79:
Title pretty much says it all - previously, UVs would only work if they were per-vertex, which is a minority of UVs.  It would also raise an error if there were no UVs.

I also added support for explicitly-authored normals to be read from USD.  This should work with per-vertex or per-face-vertex normals.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -63,6 +63,7 @@ else()
 		hdSt
         hdx
         usdLux
+        usdUtils
         pxOsd)
 endif()
         
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -22,6 +22,40 @@
 
 PXR_NAMESPACE_OPEN_SCOPE
 
+namespace {
+	template<typename OutputType>
+	void ReadPrimvar(HdSceneDelegate * sceneDelegate, SdfPath const& primId, TfToken primvarName,
+				HdPrimvarDescriptor const& primvarDescriptor, OutputType& outValue,
+				InterpolationType& outInterpolation)
+	{
+		// if outInterpolation is non-None, we assume we've already read this primvar
+		if (outInterpolation != InterpolationType::NONE)
+		{
+			return;
+		}
+		if (primvarDescriptor.name == primvarName) {
+			VtValue normalsValue = sceneDelegate->Get(primId, primvarName);
+			if (normalsValue.IsHolding<OutputType>()) {
+				switch(primvarDescriptor.interpolation)
+				{
+				case HdInterpolationVertex:
+					outInterpolation = InterpolationType::Vertex;
+					break;
+				case HdInterpolationFaceVarying:
+					outInterpolation = InterpolationType::FaceVarying;
+					break;
+				default:
+					TF_CODING_ERROR("UVs had unsupported interpolation type: %d",
+									primvarDescriptor.interpolation);
+					return;
+				}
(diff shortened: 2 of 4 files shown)


bsavery:
Adding @hshakula  to review.  He was looking at a few things textures related too.  Specifically https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/issues/24

hshakula:
Hi @elrond79, please have a look at #28.
As I can see everything that can be added by this pull request is already in master.
But by using these changes we can avoid getting the error from inside of sceneDelegate methods due to missing 'st' or 'normals' with the price of iterating through all primvars, but typically there is not so much of them so it's pretty cheap. From another side, this error is just due to limited HdSceneDelegate API as you pointed [here](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/032e619389c7c30f5b80580178ff68119a6aa206/pxr/imaging/plugin/hdRpr/mesh.cpp#L145). Actually if avoiding this error is highly desired then we can add this explicit check through HdPrimvarDescriptor's to the current master version.
Please tell me if I missed something so we could deal with it

hshakula:
Also, when sceneDelegate->Get method used, you get flattened version of primvar instead of the original. Because of this when faceVarying interpolation is specified we have to generate dummy indices array filled with [0, 1, 2, ..., numIndices] to properly map it to RPR. This causes much bigger memory usage with big assets.
Instead, the master version pulls out the original data (using sceneDelegate->SamplePrimvar) and checks if there are indices for this primvar

elrond79:
Doh - I missed your PR! Should have checked the PRs first before making my own...

You're right, these do cover almost exactly the same ground... with the differences you point out.  So I'll close this PR, and we can go from master.  However:

- You note that your version grabs the indices directly.  I can see why you'd want to do this - as you point out, it would be more efficient - but the code as you've written will fail... the reason being that the primvar funcs check for and explicitly reject any primvar with a name ending with ":indices".  So I believe it will always fail that branch, and fall through to the section which checks primvar descriptors (and effectively acts as my branch).
- Even if it DID grab the indices, I don't believe it would act as you desire. The results you get from querying the primvar (ie, SamplePrimvar, or a simple "Get") already have the indices "baked" in - they're preflattened. So trying to apply these indices to the already flattened result would give an incorrect result.  If we wish to avoid having to create and pass along "dummy" indices, then we would need to modify rprContextCreateMesh itself to have a notion of different interpolation types.

hshakula:
- It spams errors when trying to read indices but if they are present successfully returns them. [My testing assets](https://developer.apple.com/augmented-reality/quick-look) . If you tested and got USD stage that does not behave as I believe it should, please drop me asset so I would be able to debug it
- 'SamplePrimvar' returns original (not flattened) data, that's the difference from 'Get' method. Working as intended with my testing set and again if you will find the stage with buggy behavior I would appreciate if you notify me

elrond79:
Well, you're right on the first point - it does indeed return a value after printing the error.  However, you're incorrect on the second.  SamplePrimvar does return flattened data: you can see this with the attached file (it's a .usdz - you can simply rename the extension, or unpack it).

[cube_texture2.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/3408591/cube_texture2.zip)

As you can tell from the screenshot, the uvs are clearly wrong:

cube_texture2 screenshot

You can also just confirm by printing the size of the uvs returned from SamplePrimvar - you'll see it prints 24 (the number of face verts), and not 14 (the number of raw unflattened uvs in the file).

Finally - even if this did happen to work with a usd input, it should STILL not be done, as it's a detail of the UsdImaging scene delegate - other scene delegates might implement these things in a completely different way.

PR #35, opened 2019-07-18
Change usage of RprTools

elrond79:
So, this may be to open a dialog... as apparently the old way was working for people, but did not for me.

ie, previously, the code would `include "../RprTools.h"`, even though all other rpr includes didn't have the ".." - and all these files are in the same folder in my rpr install

Perhaps this is a detail which varies by OS? On my install for CentOS, all headers, as well as RprTools.h and RprTools.cpp, are all in the same folder.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -5,8 +5,8 @@
 #include "RadeonProRender_CL.h"
 #include "RadeonProRender_GL.h"
 
-#include "../RprTools.h"
-#include "../RprTools.cpp"
+#include "RprTools.h"
+#include "RprTools.cpp"
 
 #include "material.h"
 #include "materialFactory.h"


bsavery:
So when you say in the same folder as your RPR install, what do you mean?  The RPR directory should be something like this: 

And I think the include dirs are setup to look in $RPR/inc/ so to find the RprTools.h it would have to look in ../RprTools.h.  Perhaps the include directory should include $RPR and $RPR/inc though? 

Screen Shot 2019-07-17 at 6 37 00 PM

hshakula:
It does not compile on macOS (clang 10.0.1).
Also, master was successfully compiled on my CentOS 7.6 with the shipped toolchain.
Can you please try this:
```
#include <../RprTools.h>
#include <../RprTools.cpp>
```
If it will not work, let us do as @bsavery suggested including your changes

bsavery:
Thinking about this, I'd prefer to change the include path and no ../ in the include statement.

elrond79:
Ah... so, I now realize that the installation step/script that I thought was part of the official distribution was in fact just created by someone else here at Luma.  Apologies!  I'll close this PR.

bsavery:
I think the PR is valid if included with the cmake file change.  Just not urgent ;)

PR #37, opened 2019-07-18
Refactoring: Use new RPR (1.34.1) material system

hshakula:
- Remove RPRX material system usage, use plain RPR material system instead.
- Fix numerous memory leaks due to missing 'DeleteMaterial' calls.
- Force cmake to search RPR libraries only in RPR_LOCATION_LIB

### DIFF
--- a/pxr/imaging/plugin/hdRpr/materialAdapter.cpp
+++ b/pxr/imaging/plugin/hdRpr/materialAdapter.cpp
@@ -225,7 +225,7 @@ void getTextures(const  HdMaterialNetwork & materialNetwork, MaterialTextures & 
 			getParam(HdRprTokens->scale, node, param);
 			if (param.IsHolding<GfVec4f>())
 			{
-				materialNode.IsScaleEmable = true;
+				materialNode.IsScaleEnabled = true;
 				materialNode.Scale = param.Get<GfVec4f>();
 			}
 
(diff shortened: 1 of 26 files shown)


pxr/imaging/plugin/hdRpr/materialAdapter.cpp
```diff
@@ -376,27 +376,27 @@ void MaterialAdapter::PoulateUsdPreviewSurface(const MaterialParams & params, co
 		}
 		else if (paramName == HdRprTokens->metallic)
 		{
-			m_vec4fRprxParams.insert({ RPRX_UBER_MATERIAL_REFLECTION_METALNESS, VtValToVec4f(paramValue)});
```
bsavery: Is RPRX_UBER_MATERIAL_REFLECTION_METALNESS_MODE set somewhere?

bsavery:
I think we should keep this in a branch until RPR SDK 1.34.1 is released publically.  I'll test more but looks ok on a first pass.

pxr/imaging/plugin/hdRpr/materialAdapter.cpp
hshakula: Looks like no and no one used this before
RPRX_UBER_MATERIAL_REFLECTION_METALNESS_MODE is not present in RprSupport.h and RadeonProRender.h
Are you sure that such things exist?

bsavery: Sorry RPRX_UBER_MATERIAL_REFLECTION_MODE

hshakula: [here](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/5b12a652c648122c02a162cd2bc9a827bcbc195e/pxr/imaging/plugin/hdRpr/materialAdapter.cpp#L462) and [here](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/5b12a652c648122c02a162cd2bc9a827bcbc195e/pxr/imaging/plugin/hdRpr/materialAdapter.cpp#L474)

PR #38, opened 2019-07-18
fix sphereLight normalization

elrond79:
were two issues:

1. it didn't take the radius into account
2. the formula for surface area was wrong the "3/" should have been inside
   the outermost pow, not outside

### DIFF
--- a/pxr/imaging/plugin/hdRpr/sphereLight.cpp
+++ b/pxr/imaging/plugin/hdRpr/sphereLight.cpp
@@ -53,14 +53,17 @@ RprApiObject HdRprSphereLight::CreateLightMesh(std::map<TfToken, float> & params
 
 GfVec3f HdRprSphereLight::NormalizeLightColor(const GfMatrix4d & transform, std::map<TfToken, float> & params, const GfVec3f & inColor)
 {
+	// Approximating the area of a stretched ellipsoid using the Knud Thomsen formula:
+	// http://www.numericana.com/answer/ellipsoid.htm
 	constexpr const double p = 1.6075;
 	constexpr const double pinv = 1. / 1.6075;
+	const float radius = params[HdRprSphereLightTokens->radius];
 
-	const double sx = GfVec3d(transform[0][0], transform[1][0], transform[2][0]).GetLength();
-	const double sy = GfVec3d(transform[0][1], transform[1][1], transform[2][1]).GetLength();
-	const double sz = GfVec3d(transform[0][2], transform[1][2], transform[2][2]).GetLength();
+	const double sx = GfVec3d(transform[0][0], transform[1][0], transform[2][0]).GetLength() * radius;
+	const double sy = GfVec3d(transform[0][1], transform[1][1], transform[2][1]).GetLength() * radius;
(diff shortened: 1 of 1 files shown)


bsavery:
I haven't run the math on this, but isn't there a 4p term mentioned in that link?  

`S   »   4p  [ ( a^pb^p + a^pc^p + b^pc^p ) / 3 ] ^ 1/p`

elrond79:
There is, but we're just interested in the relative ratio:
surface_area(our_light) / surface_area(standard_light) - so any constant
terms will cancel out.  Or, put another way - I'm assuming we want to
standardize on a unit sphere, so any constant should be chosen such that a
sphere with radius = 1 will give a multiplicative factor of 1.

Though - come to think of it, I actually just assumed the "standard" was a
sphere with radius 1.  It might just as easily be a sphere with diameter
1... in which case we would need a corrective factor.  Is there a standard
consensus on this in the rendering community at large?

Interesting note: on the page I linked,
http://www.numericana.com/answer/ellipsoid.htm, it uses 4*p, whereas on
wikipedia (https://en.wikipedia.org/wiki/Ellipsoid#Approximate_formula), it
uses 4*pi... Not sure which one is correct, but again, for our use case, it
doesn't matter...

elrond79:
Just for completeness - cross referencing with the surface area of a sphere
- 4*pi*r^2 - the correct factor is 4*pi, not 4*p.  Wikipedia wins!

bsavery:
Yeah I was wondering about the pi factor as well.  Seemed strange there wasn't one in there ;) . 

Unless I'm misreading USDLux the default radius is 1 btw .

pxr/imaging/plugin/hdRpr/sphereLight.cpp
```diff
@@ -53,14 +53,17 @@ RprApiObject HdRprSphereLight::CreateLightMesh(std::map<TfToken, float> & params
...
 	constexpr const double p = 1.6075;
 	constexpr const double pinv = 1. / 1.6075;
+	const float radius = params[HdRprSphereLightTokens->radius];
 
-	const double sx = GfVec3d(transform[0][0], transform[1][0], transform[2][0]).GetLength();
-	const double sy = GfVec3d(transform[0][1], transform[1][1], transform[2][1]).GetLength();
-	const double sz = GfVec3d(transform[0][2], transform[1][2], transform[2][2]).GetLength();
+	const double sx = GfVec3d(transform[0][0], transform[1][0], transform[2][0]).GetLength() * radius;
+	const double sy = GfVec3d(transform[0][1], transform[1][1], transform[2][1]).GetLength() * radius;
+	const double sz = GfVec3d(transform[0][2], transform[1][2], transform[2][2]).GetLength() * radius;
```
bsavery: BTW we could also add an optimization here of course if SX == SY == SZ not a huge deal though

elrond79: Hmm - yeah, the approximation is exact when SX == SY == SZ, so we could switch over to the optimization without worrying about any "jump"...

PR #39, opened 2019-07-18
avoid attempting to get primvar indices directly

elrond79:
While attempts to get, ie, "primvars:st:indices" directly will work (after printing ominous
errors), it should be avoided because the results returned from Get and SamplePrimvar are
already flattened - and re-indexing will give incorrect results.

Also, even if it did work, the fact that indices may be stored at "primvars:st:indices" is
an implementation detail of the UsdImaging scene delegates - other scene delegates could
behave differently.

This resolves the issues discussed here:

https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/34#issuecomment-513005738

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -63,6 +63,7 @@ else()
 		hdSt
         hdx
         usdLux
+        usdUtils
         pxOsd)
 endif()
         
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -18,6 +18,8 @@
 
 #include "pxr/imaging/pxOsd/subdivTags.h"
 
+#include "pxr/usd/usdUtils/pipeline.h"
+
 PXR_NAMESPACE_OPEN_SCOPE
 
 HdRprMesh::HdRprMesh(SdfPath const & id, HdRprApiSharedPtr rprApiShared, SdfPath const & instancerId) : HdMesh(id, instancerId)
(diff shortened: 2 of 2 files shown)


hshakula:
- Could not reproduce hypothesis: 'results returned from Get and SamplePrimvar are
already flattened'. When I test cube_texture2.usdz, SamplePrimvar returns 14 element array for 'pCube1' st i.e. not flattened. 
- primvars:primvarName:indices is not a detail of UsdImagingDelegate, but strictly defined behavior of USD core - [UsdGeomPrimvar](https://graphics.pixar.com/usd/docs/api/class_usd_geom_primvar.html) Indexed primvars API

In case if we want to avoid getting indices directly, SamplePrimvar call should be replaced with 'Get'

For now, in this piece of code, only one detail of UsdImagingDelegate used - flattening of primvars. The fact that HdSceneDelegate flattens primvars in 'Get' call is not documented behavior. Actually, there is zero info 'Get' (to be precise, that's all we know about 'Get': "Returns a named value.").
And therefore, I believe that if someone would like to implement his own HdSceneDelegate, he will rely not on documentation, because it is almost empty, but on the behavior of the main delegate implementation in the USD package - UsdImagingDelegate. Because otherwise, the delegate will not work as it should with other modules that were written before

elrond79:
Ah - Ok, I see what's going on here now.  I had been using the latest dev branch of USD - when I tested with 19.07, I saw the same results you're seeing.

The difference comes from this commit: https://github.com/PixarAnimationStudios/USD/commit/1e081487bdcc82922bc941cb050a34edb66b1c2a

This fixed a bug with SamplePrimvar, that made it return non-flattened results... thus the source of confusion.  Regardless, the answer is clearly that we should NOT be relying on getting non-flattened results from SamplePrimvar.

As for relying on things in UsdImagingDelegate - you do have a point, in that documentation IS skimpy, and so sometimes you have to rely on the behavior of UsdImagingDelegate as a reference.  This is certainly what we did at times when implementing our Maya scene delegate.  And "HdSceneDelegate::Get" is, as you point out, a bit of a wild west - it's never clearly defined what values must be supported, or what form they should take, etc.  However, I think it's safe to say that if even UsdImagingDelegate explicitly errors or warns us against a behavior - as it does with attempts to retrieve "primvars:st:indices" directly - that this is NOT behavior we should be relying on.

Finally - when I said that the use the indices attribute was an "implementation detail of the UsdImaging scene delegate", I did not mean that the code defining this behavior was defined within the UsdImagingDelegate class - instead, I meant that it was a detail of the USD file format, and therefore, from a render delegate perspective, can only be relied on to exist when a particular rprim is backed by the UsdImagingDelegate.  In general, pretty much everything in UsdGeom falls under this category - everything there is implementing details of the USD file format, and will not necessarily be the same when accessed through other scene delegates.

hshakula:
After some days of sitting with this question, I am on your side. In perspective, we may have more problems than benefits. And with the current approach, vertex interpolated indexed primvars does not work correctly if I'm getting it right.
But it does not mean what I can say more, because as we can see, in the next version of USD this question will be outdated anyway.

PR #41, opened 2019-07-25
Integrate hybrid rendering backend plugin

hshakula:


pxr/imaging/plugin/hdRpr/materialFactory.cpp
```diff
@@ -254,6 +261,15 @@ RprApiMaterial* RprMaterialFactory::CreateMaterial(EMaterialType type, const Mat
             continue;
         }
 
+        // SIGGRAPH HACK: Fix for models from Apple AR quick look gallery.
+        //   Of all the available ways to load images, none of them gives the opportunity to
+        //   get the gamma of the image and does not convert the image to linear space
```
bsavery: Next version of RPR will have something for loading decoded image from memory fyi

hshakula: Yeap, rprContextCreateImageFromFileMemory

hshakula: Next version is going to solve gamma issue as well

PR #42, opened 2019-08-23
Add ability to build as Houdini plugin

hshakula:
Inside you will find:
- Established cmake schema to build plugin either as usdview plugin or Houdini plugin by setting `PXR_BUILD_AS_HOUDINI_PLUGIN` with an appropriate value
- Basic implementation for Hydra AOV system, but disabled yet
- Ability to enable RPR tracing using environment variables


CMakeLists.txt
```diff
@@ -12,6 +12,30 @@ set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/defaults
                       ${CMAKE_SOURCE_DIR}/cmake/macros)
 
 include(Options)
+
+if(PXR_BUILD_AS_HOUDINI_PLUGIN)
```
bsavery: We should probably namespace this as RPR_BUILD_AS_HOUDINI_PLUGIN.  PXR-* is for pixar/USD

README.md
```diff
@@ -1,86 +1,95 @@
-AMD Radeon ProRender delegate for USD (hdRpr)
+AMD Radeon ProRender rendering plugin for USD Hydra
```
bsavery: Technically this is a "render delegate".  Maybe that would be a good way instead of plugin.

CMakeLists.txt
hshakula: I added `PXR_BUILD_AS_HOUDINI_PLUGIN` guided by `PXR_ENABLE_RIF_SUPPORT`. Perhaps, in that case, we should consider renaming this variable as well?

README.md
hshakula: It's actually a delegate, but technically it's a plugin. So maybe let's name it as USD team [names](https://graphics.pixar.com/usd/docs/api/hd_embree_page_front.html) embree delegate? What do you think?
> HdEmbree : Embree-based hydra renderer plugin.

CMakeLists.txt
bsavery: Yes, that should be renamed as well...

hshakula:
I'm closing this PR due to its size and then going to create smaller PRs

PR #54, opened 2019-10-02
Add support for orthographic camera

hshakula:


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1065,6 +1053,49 @@ class HdRprApiImpl {
...
+        auto iwvm = m_cameraViewMatrix.GetInverse();
+        auto& wvm = m_cameraViewMatrix;
+
+        GfVec3f eye(iwvm[3][0], iwvm[3][1], iwvm[3][2]);
+        GfVec3f up(wvm[0][1], wvm[1][1], wvm[2][1]);
+        GfVec3f n(wvm[0][2], wvm[1][2], wvm[2][2]);
+        GfVec3f at(eye - n);
+        RPR_ERROR_CHECK(rprCameraLookAt(m_camera, eye[0], eye[1], eye[2], at[0], at[1], at[2], up[0], up[1], up[2]), "Fail to set camera Look At");
+
+        bool isOrthographic = round(m_cameraProjectionMatrix[3][3]) == 1.0;
```
bsavery: This is how we detect if it's orthographic?  There's no flag from USD?

hshakula: Yes, there is no flags for it. I spied this solution in hdEmbree

hshakula: HdCamera has GetProjection method, but when I discovered that we don't support ortho projection, this method always returned perspective, even if orthographic expected. At least I was seeing such behavior in Houdini

PR #56, opened 2019-10-02
Remove any headers from the public interface

hshakula:
* Plugins should not expose any headers with implementation details

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -91,7 +91,7 @@ pxr_plugin(hdRpr
 		
 		${OptIncludeDir}
 				
-    PUBLIC_CLASSES
+    PRIVATE_CLASSES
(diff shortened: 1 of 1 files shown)


pxr/imaging/plugin/hdRpr/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ pxr_plugin(hdRpr
 		
 		${OptIncludeDir}
 				
-    PUBLIC_CLASSES
+    PRIVATE_CLASSES
 		rendererPlugin
         renderDelegate
```
bsavery: is the indentation weird here?

hshakula: I try to not fix code style issues in such commits. Otherwise, it would be much bigger commits

hshakula: I will create one big PR with code style fixes when we are done with all current PRs

PR #61, opened 2019-10-03
Refactor lifetime management of HdRprApi objects (mesh, curve, etc)

hshakula:
Enable automatic release of rpr objects with the functionality of invoking custom clean up callable objects on release (like detaching env light or detaching shape from the scene).

Such refactoring would prevent all possible memory leaks and dangling pointers in RPR state machine.

This changes will fix such existing memory leaks:
* lightBase material
* Environment light
    > When you remove dome light from the scene it would stay there anyway, you can see it easily in Houdini
* maybe another issue I don't know about

#### Improvements

I would like to make RprApiObject more type-safe, but the easiest way to accomplish it is to make rpr_* types as type-safe. For now, rpr types defined like this:
```
typedef void * rpr_context;
typedef void * rpr_camera;
```
but from type standpoint rpr_context == rpr_camera. You can easily create a rpr_camera variable and pass it to function that expects rpr_context, you will not have any compiler error or even warnings.


bsavery:
In essence this is ok.  My one question really doesn't block the PR.  But this part of the idea of what we are trying to accomplish with the wrapper api that we discussed.  We should talk about how to incorporate these ideas into that to share across various plugins.

hshakula:
I look forward to Richard posting his changes to the repository so that we can work together on improving the wrapper

PR #63, opened 2019-10-04
Update to RPR SDK 1.34.3

hshakula:
RPR 1.34.3 added new function - `rprFrameBufferFillWithColor` - that will allow me to fix a few bugs we have. So as previously we had no ability to fill framebuffer with custom color, we always had to fill it with zero using rprFrameBufferClear. It causes such inconsistencies for pixels that were not hit by ray:
* WorldCoordinate AOV (used for ndc depth calculation) - invalid depth value
* ObjectID AOV - conform to the first index, but actually should be some sort of invalid index (e.g. -1)


hshakula:
Not ready to merge. Few issues with materials in this version of RPR
[images2.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/3690957/images2.zip)
[video.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/3690965/video.zip)
Looks like normal and roughness mapping get broken
Normal mapping is definitely broken, you can see it from video via flat shading on mesh

bsavery:
1.34.2 look correct though no?
I'm going to push some more changes.

hshakula:
I've tested this release https://github.com/Radeon-Pro/RadeonProRenderSDK/archive/v1.34.2.zip on Windows, it gives me the same results as the latest release

bsavery:
My preference is to merge this even with material issues and update when possible.  @hshakula thoughts?

hshakula:
Well, in the current situation, from hdRpr standpoint more correct solution is to stay up to date with the latest RPR SDK, even if has some bugs. IMO you need to push the idea with RPR SDK submodule, ofc if you have not already done so

PR #65, opened 2019-10-07
Fix cmake project configuration on Ubuntu

hshakula:


cmake/defaults/CXXDefaults.cmake
```diff
@@ -102,3 +102,10 @@ endif()
 #
 # The long term plan is to adhere to the required behavior.
 include_directories(SYSTEM ${PYTHON_INCLUDE_DIR})
+
+# hdRpr changes
+if (APPLE)
+    set(CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS "${CMAKE_SHARED_LIBRARY_CREATE_CXX_FLAGS} -undefined dynamic_lookup")
```
bsavery: This should be removed?  We removed for apple in other PR

cmake/macros/PlatformIntrospection.cmake
```diff
@@ -0,0 +1,122 @@
+# Returns a lowercased version of a given lsb_release field.
```
bsavery: Is this code taken from an example somewhere?

cmake/modules/FindRif.cmake
```diff
@@ -1,42 +1,23 @@
...
+
+if(APPLE)
+    SET_RIF_VARIABLES(OSX)
+elseif(WIN32)
+    SET_RIF_VARIABLES(Windows)
+elseif(PLATFORM STREQUAL "ubuntu18.04")
+    SET_RIF_VARIABLES(Ubuntu18)
+elseif(PLATFORM STREQUAL "ubuntu16.04")
+    SET_RIF_VARIABLES(Ubuntu)
+elseif(PLATFORM STREQUAL "centos7")
```
bsavery: so this is ok.  Let's say I'm on centos 8 and maybe want to force it to use the centos7 build.  How would I do this?

cmake/macros/PlatformIntrospection.cmake
hshakula: Yep, taken from here https://github.com/couchbase/tlm
As I've checked there is no licensing. Can we use it?

cmake/modules/FindRif.cmake
hshakula: We can make PLATFORM variable as an option and let the user specify it otherwise calculate it automatically as it is now. In that case, we should consider renaming it

cmake/macros/PlatformIntrospection.cmake
bsavery: ok

cmake/modules/FindRif.cmake
bsavery: Ok please do that.

PR #66, opened 2019-10-07
Fix boost and glew cmake configuration for Houdini plugin

hshakula:


bsavery:
It might be worth splitting the boost part out.  This part seems to work on macOS.

hshakula:
Without handling GLEW library there is no point in the correct handling of "boost"

hshakula:
Houdini 18.0.252 resolved this issue. GLEW headers have been added.
Tested on mac and windows

Issue #67, opened 2019-10-07
Issues on macOS build


bsavery:
1.  remove -undefined dynamic_lookup flags that hide missing symbols
2.  This leads to errors that symbols like rifCreateContextFromOpenClContext missing on macOS.  Possible solutions are 
     a:  ifdef around calls in code (ok for now)
     b: RIF implements errors for throwing errors if not supported
3.  Use RPR_CREATION_FLAGS_ENABLE_METAL on metal.  CL has issues.

PR #70, opened 2019-10-07
Fixes needed for macOS

bsavery:
macOS should always use METAL.  Also remove -undefined dynamic_lookup as it hid link errors.  Also #ifdef out some calls that are not defined on macOS.

Fixes https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/issues/67


hshakula:
Tested in the plugin for usdview - it's successfully compiled and has valid render output

PR #71, opened 2019-10-08
Fix hdRpr's install name of RPR library

hshakula:
It looks like previously RPR library had broken rpath and so was added this workaround. But now it's redundant due to correct rpath in RPR SDK and even more, it does not allow us to distribute build among users due to hardcoded install name of RPR library

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -138,16 +138,6 @@ if(WIN32 OR LINUX)
 	add_definitions(-DUSE_GL_INTEROP)
 endif(WIN32 OR LINUX)
 
-if (APPLE)
-    add_custom_command(TARGET hdRpr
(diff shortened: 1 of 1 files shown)


bsavery:
I still see a bad path in libTahoe... 

> otool -L ~/Downloads/RadeonProRenderSDK-1.34.1/RadeonProRender/binMacOS/libTahoe64.dylib         
> /Users/bsavery/Downloads/RadeonProRenderSDK-1.34.1/RadeonProRender/binMacOS/libTahoe64.dylib:
> 	../dist/release/bin/x86_64/libTahoe64.dylib (compatibility version 0.0.0, current version 0.0.0)
> .... 
> 

Should we address this?

hshakula:
We do not care about it due to the way of how we use libTahoe - rprRegisterPlugin

hshakula:
If we would link libTahoe to hdRpr it would be the problem, but there is no reason for it

PR #81, opened 2019-10-10
Add sampling controls

bsavery:
This checkin adds some sampling controls.  Namely min samples, max samples, and variance.

We set is_converged on the render buffer and the render pass based on the criteria
 is samples > max samples or are all pixels converged to variance threshold.

Open TODOs:
How do we mark buffers and render pass as not converged and set iterations to 0 when changes happen?
How do we set defaults here?
Expose in Menu?

### DIFF
--- a/pxr/imaging/plugin/hdRpr/config.cpp
+++ b/pxr/imaging/plugin/hdRpr/config.cpp
@@ -61,6 +61,42 @@ bool HdRprConfig::IsDenoisingEnabled() const {
     return m_prefData.m_enableDenoising;
 }
 
+void HdRprConfig::SetMinSamples(int minSamples) {
+    if (m_prefData.m_minSamples != minSamples) {
+        m_prefData.m_minSamples = minSamples;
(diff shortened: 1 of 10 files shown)


hshakula:
> How do we mark buffers and render pass as not converged

What about calling HdRprRenderBuffer::SetConverged all the time in renderPass - even if isConverged is false. So that its value is always up to date

> set iterations to 0 when changes happen

When we clear our framebuffers, in HdRprApiImpl::Update inside of
```
if (m_dirtyFlags & ChangeTracker::DirtyScene)
```

> How do we set defaults here?

Do you mean default values for minSample, maxSamples and variance in HdRprConfig?

> Expose in Menu?

As I see you have already added it to RenderSettings, so it should be exposed to usdview UI in HydraSettings. We can expose these controls to Houdini's UI via [.ds file](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/houdini/HdRprPlugin_Viewport.ds)

PR #87, opened 2019-10-16
OpenVDB support

minghao-AMD:


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -470,35 +472,39 @@ class HdRprApiImpl {
         });
     }
 
-    RprApiObjectPtr CreateHeterVolume(const VtArray<float>& gridDencityData, const VtArray<size_t>& indexesDencity, const VtArray<float>& gridAlbedoData, const VtArray<unsigned int>& indexesAlbedo, const GfVec3i& grigSize) {
+	RprApiObjectPtr CreateHeterVolume(const std::vector<uint32_t>& densityGridOnIndices, const std::vector<float>& densityGridOnValueIndices, const std::vector<float>& densityGridValues,
```
bsavery: Let's rename this to CreateHeteroVolume

pxr/imaging/plugin/hdRpr/volume.cpp
```diff
@@ -20,9 +23,9 @@ TF_DEFINE_PRIVATE_TOKENS(
 
 namespace
 {
-	const float defaultDencity = 1.f;
+	const float defaultDensity = 1.f;
 	const float defaultAlbedo = 1.f;
-	GfVec4f defaultColor = GfVec4f(35.0f, 35.0f, 35.f, 1.0f);
+	GfVec4f defaultColor = GfVec4f(10.0f, 0.0f, 0.0f, 1.0f);
```
bsavery: Is there a significance to this color?  Or is this just to test?  Also it's somewhat odd to have albedos > 1.0

pxr/imaging/plugin/hdRpr/volume.cpp
```diff
@@ -348,20 +393,40 @@ void HdRprVolume::Sync(
...
-			else if (baseGrid->isType<openvdb::Vec3IGrid>())	fillGridOnWithDefaultColor< openvdb::Vec3IGrid::Ptr, openvdb::Vec3IGrid::ValueOnIter >(openvdb::gridPtrCast<openvdb::Vec3IGrid>(baseGrid), grid, idx, gridOnBBSize);
-			else if (baseGrid->isType<openvdb::Vec3SGrid>())	fillGridOnWithDefaultColor< openvdb::Vec3SGrid::Ptr, openvdb::Vec3SGrid::ValueOnIter >(openvdb::gridPtrCast<openvdb::Vec3SGrid>(baseGrid), grid, idx, gridOnBBSize);
-			else if (baseGrid->isType<openvdb::Vec3DGrid>())	fillGridOnWithDefaultColor< openvdb::Vec3DGrid::Ptr, openvdb::Vec3DGrid::ValueOnIter >(openvdb::gridPtrCast<openvdb::Vec3DGrid>(baseGrid), grid, idx, gridOnBBSize);
-			else if (baseGrid->isType<openvdb::StringGrid>())	fillGridOnWithDefaultColor< openvdb::StringGrid::Ptr, openvdb::StringGrid::ValueOnIter >(openvdb::gridPtrCast<openvdb::StringGrid>(baseGrid), grid, idx, gridOnBBSize);
-			*/
+			if (baseGrid->isType<openvdb::BoolGrid>())
+				voxelSize = voxelSize;
+				//fillGridOnWithDefaultColor< openvdb::BoolGrid::Ptr, openvdb::BoolGrid::ValueOnIter >(openvdb::gridPtrCast<openvdb::BoolGrid>(baseGrid), grid, idx, gridOnBBSize);
+			else if (baseGrid->isType<openvdb::FloatGrid>())
```
bsavery: Would this make more sense to do 
if (baseGrid->isType<openvdb::FloatGrid>()) {
    ... 
} else {
    voxelSize = voxelSize;
}

Unless the intention is to fill in more here?

pxr/imaging/plugin/hdRpr/volume.cpp
```diff
@@ -154,17 +159,59 @@ openvdb::CoordBBox computeGrigOnBB(TPtr grid)
...
 {
-	openvdb::CoordBBox gridOnBB = computeGrigOnBB< TPtr, TValueOnIt>(grid);
+	openvdb::CoordBBox gridOnBB = computeGridOnBB< TPtr, TValueOnIt>(grid);
 	out_gridOnBBSize = gridOnBB.max() - gridOnBB.min();
 
 	for (TValueOnIt iter = grid->beginValueOn(); iter; ++iter) {
-		out_grid.push_back(defaultColor);
-		out_idx.push_back(coordToIndex(iter.getCoord(), gridOnBB, out_gridOnBBSize));
+		out_density_grid.push_back(grid->getAccessor().GetValue(iter.getCoord()));
+		out_albedo_grid.push_back(defaultColor);
```
hshakula: `out_albedo_grid` - vector of floats, `defaultColor` - GfVec4f

pxr/imaging/plugin/hdRpr/volume.cpp
```diff
@@ -154,17 +159,59 @@ openvdb::CoordBBox computeGrigOnBB(TPtr grid)
...
+		size_t idx = coordToIndex(iter.getCoord(), gridOnBB, out_gridOnBBSize);
+		out_density_idx.push_back(idx);
+		out_albedo_idx.push_back(idx);
+	}
+}
+
+template <class TGrid>
+void ReadDensityFromGrid(openvdb::GridBase::Ptr inGrid, std::vector<uint32_t> &outDensityGridOnIndices, std::vector<float> &outDensityGridOnValueIndices, std::vector<float> &outDensityGridValues)
+{
+	TGrid::Ptr grid = openvdb::gridPtrCast<TGrid>(inGrid);
```
hshakula: type name dependent on `TGrid` requires `typename` specifier
[doc](https://en.cppreference.com/w/cpp/language/dependent_name)

pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -5,6 +5,8 @@
 #include "rifcpp/rifImage.h"
 
 #include <RadeonProRender.h>
+#include <math/matrix.h>
+#include <math/mathutils.h>
```
hshakula: Is it necessary?

minghao-AMD: hmm, weird. I think I had to add it to get rid of some compile errors. But now it also compiles without these two includes. I'll remove them.

pxr/imaging/plugin/hdRpr/volume.cpp
```diff
@@ -1,6 +1,9 @@
 #include "volume.h"
-
+#pragma warning(push)
+#pragma warning(disable:4146)
 #include "openvdb/openvdb.h"
+#pragma warning(pop)
```
hshakula: I am in favor of removing this. So as it's MSVC only solution and actually we should not care about warnings coming from external headers. Maybe we could contribute to openvdb repository to fix this warning, ofc if it's possible

pxr/imaging/plugin/hdRpr/volume.cpp
```diff
@@ -15,162 +18,107 @@ TF_DEFINE_PRIVATE_TOKENS(
...
 		}
 	}
 }
 
-template <class TPtr, class TValueOnIt>
-openvdb::CoordBBox computeGrigOnBB(TPtr grid)
+void findTemperatureOffsetAndScale(HdSceneDelegate* sceneDelegate, const SdfPath & id, float &outOffset, float &outScale)
 {
-	if (grid->empty())
+	for (auto it : sceneDelegate->GetVolumeFieldDescriptors(id))
```
hshakula: Could you please query VolumeFieldDescriptors once at Sync and pass them to all those "find*something" functions? This will allow us to avoid the unnecessary copy of the same vector every time you try to "find" something

pxr/imaging/plugin/hdRpr/volume.cpp
minghao-AMD: code refactored in latest commit

pxr/imaging/plugin/hdRpr/volume.cpp
minghao-AMD: This function removed in latest commit.

pxr/imaging/plugin/hdRpr/volume.cpp
minghao-AMD: Removed template from this function in latest commit.

pxr/imaging/plugin/hdRpr/volume.cpp
minghao-AMD: Removed.

pxr/imaging/plugin/hdRpr/volume.cpp
minghao-AMD: Wiil do.

bsavery:
@hshakula can you give final review?

hshakula:
I've tested it with varying field's **filepath** attribute. It renders only the first volume sample. Probably, we have to move the management of fields from HdRprVolume to HdRprField.

Also, I've tested in Houdini, Solaris/Volumes.hip. It does not work as well. But there is another issue - **filepath** we get from sceneDelegate comes in such form:
```
op:/obj/geo1/vdbfrompolygons1.sop:SDF_FORMAT_ARGS:globalauthortimesamples=0&t=0
```
which obviously is not a file path. I will try to get more info about this one on sidefx forum

hshakula:
Hi @minghao-AMD, could you please explain why do you compute color from the temperature grid instead of using color from the color grid? As I can see you decide whether colors required or not from querying if a color field exists, but then this particular field is not used

minghao-AMD:
The colored volume assets I used for testing (from https://www.openvdb.org/download/) only had temperature field, hence I could only extract color from temperature. I agree that color should also be considered but I just never had the chance to really test on any real assets, hence didn't connect them.

bsavery:
Hi @minghao-AMD  we are including this in the other change to include Houdini support.  Thanks for the work!

PR #88, opened 2019-10-17
Add simple image cache

hshakula:
* Minimize material memory usage
* Speed up material load time

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -112,6 +112,7 @@ pxr_plugin(hdRpr
 		sphereLight
         renderBuffer
 		basisCurves
+        imageCache
 		
 		${OptClass}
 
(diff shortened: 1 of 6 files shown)


bsavery:
Couple thoughts here:

1.  How do we mark an image in the cache as dirty?  Say if an image is updated on disk?  Is that dirty flag updated in USD?
2.  Does this handle in memory images and RAT textures?
3.  Somewhat tangental, do we do the same thing for instancers?  I.e. get the existing rpr_shape for the master instance rather than creating new each time?

hshakula:
1. As far as I know, USD do not track image file updates. Point me to this functionality if I'm wrong (I didn't manage to find anything regarding this)
2. This cache handle images incorporated inside of .usdz if you meant it. And it supports RAT through rpr::Image constructor.
3. For instancers we have the opposite situation - each rpr_shape sets its instances

bsavery:
1.  Well if the images are on disk I suppose we could check the timestamp on disk when calling GetImage to see if dirty.  I'm not sure otherwise 
2.  No I answered my own question.  Nevermind.  
3.  Ok

bsavery:
#1 would be a further optimization we could do in the future I suppose.

hshakula:
I created #91  so as not to forget about it

Issue #95, opened 2019-10-18
Opening another stage through usdview's UI does not reset iteration count


hshakula:
### Steps to reproduce
1. Open some stage in usdview
2. Render until max samples are reached
3. Open another stage using `File - Open`
4. The new stage is incorrectly rendered

PR #98, opened 2019-10-21
Fix iteration count initialization

hshakula:
After the first hdRpr initialization, HdRprConfig's sampling attributes were marked as clean. But when you reinitialize hdRpr, HdRprConfig's sampling is no longer dirty that's why the number of max samples was never queried again until you set a new value for one of the sampling attributes in HdRprConfig. Thus the number of max samples was always 0. This behavior was mentioned in #95

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -929,9 +929,9 @@ public:
         preferences.ResetDirty();
     }
 
-    void UpdateSampling() {
+    void UpdateSampling(bool force = false) {
         auto& preferences = HdRprConfig::GetInstance();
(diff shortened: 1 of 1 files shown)


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1076,7 +1076,7 @@ class HdRprApiImpl {
         RecursiveLockGuard rprLock(g_rprAccessMutex);
 
         // return Converged if max samples is reached
-        if (m_iter > m_maxSamples) {
+        if (m_iter >= m_maxSamples) {
```
hshakula: If we set max samples to 256 and we have already done 256 iterations we should stop render

PR #103, opened 2019-10-22
Use opacity map to define transparency map

hshakula:
Motivated by the [principled node](https://github.com/Radeon-Pro/RadeonProRenderBlenderAddon/blob/master/src/rprblender/nodes/blender_nodes.py#L827) in blender plugin

### DIFF
--- a/pxr/imaging/plugin/hdRpr/materialAdapter.cpp
+++ b/pxr/imaging/plugin/hdRpr/materialAdapter.cpp
@@ -426,7 +426,7 @@ void MaterialAdapter::PopulateUsdPreviewSurface(const MaterialParams & params, c
 		else if (paramName == HdRprTokens->opacity)
 		{
 			materialTexture.IsOneMinusSrcColor = true;
-			m_texRpr.insert({ RPR_UBER_MATERIAL_INPUT_REFRACTION_WEIGHT, materialTexture });
+			m_texRpr.insert({ RPR_UBER_MATERIAL_INPUT_TRANSPARENCY, materialTexture });
 		}
 		else if (paramName == HdRprTokens->normal)
 		{


bsavery:
So.  I'm not sure this is correct: 
opacity - float - 1.0
When opacity is 1.0 then the gprim is fully opaque, if it is smaller than 1.0 then the prim is translucent, when it is 0 the gprim is transparent.  Note that even a fully transparent object still receives lighting as, for example, perfectly clear glass still has a specular response.
opacityThreshold - float - 0.0
The opacityThreshold input is useful for creating geometric cut-outs based on the opacity input. A value of 0.0 indicates that no masking is applied to the opacity input, while a value greater than 0.0 indicates that rendering of the surface is limited to the areas where the opacity is greater than that value. A classic use of opacityThreshold is to create a leaf from an opacity input texture, in that case the threshold determines the parts of the opacity texture that will be fully transparent and not receive lighting. Note that when opacityThreshold is greater than zero, then opacity modulates the presence of the surface, rather than its transparency - pathtracers might implement this as allowing ((1 - opacity) * 100) % of the rays that do intersect the object to instead pass through it unhindered, and rasterizers may interpret opacity as pixel coverage.  Thus, opacityThreshold serves as a switch for how the opacity input is interpreted; this "translucent or masked" behavior is common in engines and renderers, and makes the UsdPreviewSurface easier to interchange.  It does imply, however, that it is not possible to faithfully recreate a glassy/translucent material that also provides an opacity-based mask... so no single-polygon glass leaves.
ior - float - 1.5
Index of Refraction to be used for translucent objects.

https://graphics.pixar.com/usd/docs/UsdPreviewSurface-Proposal.html
According to the USD preview surface spec it does refraction (with IOR) when transparency is on.  

I think instead you want to set diffuse weight lower when opacity goes down

hshakula:
Okay, I will give it a try

hshakula:
So you are proposing to change this:
```
materialTexture.IsOneMinusSrcColor = true;
m_texRpr.insert({ RPR_UBER_MATERIAL_INPUT_TRANSPARENCY, materialTexture });
```
to
```
m_texRpr.insert({ RPR_UBER_MATERIAL_INPUT_DIFFUSE_WEIGHT, materialTexture });
materialTexture.IsOneMinusSrcColor = true;
m_texRpr.insert({ RPR_UBER_MATERIAL_INPUT_REFRACTION_WEIGHT, materialTexture });
```
So that opacity map will correspond to diffuse weight, and 1.0 - opacity is going to be refraction weight, right?

bsavery:
yes

hshakula:
Outdated

PR #104, opened 2019-10-22
RenderSettings refactoring

hshakula:
* Move all logic related to initialization and change of the render settings to HdRprConfig class
* Automatically generate HdRprConfig class and HdRprPlugin_Viewport.ds files accordingly to current render settings definition
* Add render setting to control ray bouncing, raycast epsilon and radiance clamping


pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
```diff
@@ -0,0 +1,530 @@
...
+                'help': 'The Glossy Refraction Ray Depth parameter is similar to the Refraction Ray Depth. The difference is that it is aimed to work with matte refractive materials, such as semi-frosted glass.',
+                'defaultValue': 5,
+                'minValue': 2,
+                'maxValue': 50
+            },
+            {
+                'name': 'maxRayDepthShadow',
+                'ui_name': 'Shadow Ray Depth',
+                'help': 'Controls the accuracy of shadows cast by transparent objects. It defines the maximum number of surfaces that a light ray can encounter on its way causing these surfaces to cast shadows.',
+                'defaultValue': 5,
```
bsavery: I also think this needs to be like 2

pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
```diff
@@ -0,0 +1,530 @@
...
+                'defaultValue': 8,
+                'minValue': 2,
+                'maxValue': 50
+            },
+            {
+                'name': 'maxRayDepthDiffuse',
+                'ui_name': 'Diffuse Ray Depth',
+                'help': 'The maximum number of times that a light ray can be bounced off diffuse surfaces.',
+                'defaultValue': 3,
+                'minValue': 2,
```
bsavery: I think it's valid to have a min value of 1

hshakula: easy, I just took values from Maya plugin

Issue #93, opened 2019-10-18
Broken materials in tv_retro from AR Quick Look Gallery


hshakula:
RPR render:
![tv_retro_bug](https://user-images.githubusercontent.com/22181845/67078273-a4484200-f199-11e9-990a-bc71b441ac38.PNG)

Should look like this:
![image](https://user-images.githubusercontent.com/22181845/67078301-b3c78b00-f199-11e9-8966-65337d3f0ed3.png)

hshakula:
Maybe this is due to the fact that we set an emission weight to 1.0 for the whole material? Probably we should set it to only these parts where emission is non-zero

hshakula:
Fixed by #116

PR #111, opened 2019-10-28
Material fixes

hshakula:
* Fix translucent layer: `diffuse_weight = opacity` and `refraction_weight = 1 - opacity`
* Set fallback values for absent parameters
* Set refraction color to white
* Handle left-hand oriented meshes


hshakula:
There are still two things in TODO:
- Handling opacityThreshold
- Fix for semi-emissive semi-diffuse material (#93)

PR #121, opened 2019-11-02
Fix crash on resize of HdRprRenderBuffer

hshakula:
* Avoid potential crashes due to incorrect usage of HdRenderBuffer API by external code. According to my observations, HdRenderBuffer might be reallocated while in a mapped state. Also, after reallocation HdRprRenderBuffer's internal buffer used to store resolved image has been in an incorrect state - it didn't match new width and height - until first Resolve call.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/renderBuffer.cpp
+++ b/pxr/imaging/plugin/hdRpr/renderBuffer.cpp
@@ -14,10 +14,8 @@ TF_DEFINE_PRIVATE_TOKENS(
     (aov_primvars_st)
 );
 
-HdRprRenderBuffer::HdRprRenderBuffer(SdfPath const & id, HdRprApiSharedPtr rprApi) : HdRenderBuffer(id), m_numMappers(0)
-{
-    if (!rprApi)
-    {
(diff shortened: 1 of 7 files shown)


pxr/imaging/plugin/hdRpr/renderBuffer.cpp
```diff
@@ -26,35 +24,18 @@ HdRprRenderBuffer::HdRprRenderBuffer(SdfPath const & id, HdRprApiSharedPtr rprAp
...
-    {
+    } else if (idName == _tokens->aov_primId) {
         m_aovName = HdRprAovTokens->primId;
-        m_format = HdFormat::HdFormatUNorm8Vec4;
-    }
-    else if (idName == _tokens->aov_primvars_st)
-    {
+    } else if (idName == _tokens->aov_primvars_st) {
         m_aovName = HdRprAovTokens->primvarsSt;
-        m_format = HdFormat::HdFormatFloat32Vec3;
```
hshakula: Let the user of HdRenderBuffer decide which format to use for particular AOV

PR #122, opened 2019-11-02
Add disk light support

hshakula:


pxr/imaging/plugin/hdRpr/diskLight.cpp
```diff
@@ -0,0 +1,58 @@
...
+        return nullptr;
+    }
+
+    return rprApi->CreateDiskLightMesh(params[HdLightTokens->radius]);
+}
+
+GfVec3f HdRprDiskLight::NormalizeLightColor(const GfMatrix4d& transform, std::map<TfToken, float>& params, const GfVec3f& inColor) {
+    float radius = params[HdLightTokens->radius];
+    if (radius <= 0.0f) {
+        radius = 1.0f;
```
bsavery: Should this be radius = -radius?

bsavery: Rather I'm saying any -radius light should be automatically treated as the negative of the negative radius

hshakula: Maybe. This logic is a copypaste from existing logic in sphere light. If we want to take the absolute value of the radius here then for sphere light we should the same

PR #123, opened 2019-11-02
Fix color banding issues

hshakula:
HdRenderDelegate is setting up descriptors for each AOV type. Instead of using uint8 images for color output use float to allow Hydra user to apply any postprocessing wanted. Internal RPR buffers anyway always are in float format. Now we will save time on conversion to uint8


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -751,23 +750,6 @@ class HdRprApiImpl {
...
-            if (RPR_ERROR_CHECK(rprContextCreatePostEffect(m_rprContext->GetHandle(), RPR_POST_EFFECT_TONE_MAP, &tonemap), "Fail to create post effect")) return;
-            m_tonemap = RprApiObject::Wrap(tonemap);
-
-            if (RPR_ERROR_CHECK(rprContextAttachPostEffect(m_rprContext->GetHandle(), tonemap), "Fail to attach posteffect")) return;
-            m_tonemap->AttachOnReleaseAction(RprApiObjectActionTokens->attach, [this](void* tonemap) {
-                rprContextDetachPostEffect(m_rprContext->GetHandle(), tonemap);
-            });
-        }
-    }
-
```
hshakula: We should not apply implicit tone mapping. We could consider exposing a particular render setting to allow a user to select if he wants to apply tone mapping or not but in many cases, a user will want to get strict linear color without any color correction for correct compositing.

bsavery: Hmmm.  I would agree for final rendering.  For the viewport though the app may do it's own post or we could do it.

PR #128, opened 2019-11-05
Introduce basic interactivity support

hshakula:
When an interactive mode is requested, set max ray depth value to the interactive one.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
+++ b/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
@@ -121,6 +121,14 @@ render_setting_categories = [
                 'defaultValue': 0.0,
                 'minValue': 0.0,
                 'maxValue': 1e6
+            },
+            {
+                'name': 'interactiveMaxRayDepth',
+                'ui_name': 'Interactive Max Ray Depth',
+                'help': 'Controls value of \'Max Ray Depth\' in interactive mode.',
+                'defaultValue': 2,
+                'minValue': 1,
+                'maxValue': 50
             }
         ]
     }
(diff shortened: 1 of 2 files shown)


bsavery:
So houdini has a setting "Interactive mode?" Or is this strictly for our plugin?

pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
```diff
@@ -121,6 +121,14 @@
                 'defaultValue': 0.0,
                 'minValue': 0.0,
                 'maxValue': 1e6
+            },
+            {
+                'name': 'interactiveMaxRayDepth',
+                'ui_name': 'Interactive Max Ray Depth',
+                'help': 'Controls value of \'Max Ray Depth\' in interactive mode.',
+                'defaultValue': 2,
```
bsavery: Max depth 2 is very low for any glass IMO.

hshakula:
When you move camera, Houdini set render setting “houdini:interactive” to some string which do not equal “normal”. As soon as user stops moving camera, Houdini sets this render setting to “normal”. So all values other than “normal” interpreted as enabled interactive mode. This interactive mode needed for one particular reason - speed up frame render time for easier camera transform adjustment.

pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
hshakula: So as it’s only active when you move the camera it’s ok, but if it’s a problem for the user exactly for that I expose render setting to control interactive ray depth instead of using constant one

bsavery:
I see.  That makes sense.  In that case we should also set "RPR_CONTEXT_PREVIEW" = 1 which pixelates things for the first few iterations.

hshakula:
Sure! I didn’t know about such thing. Would you mind adding it?

PR #131, opened 2019-11-07
Implement distantLight support

hshakula:


pxr/imaging/plugin/hdRpr/distantLight.cpp
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        if (!m_rprLight) {
+            m_rprLight = rprApi->CreateDirectionalLight();
+            newLight = true;
+        }
+
+        float angle = sceneDelegate->GetLightParamValue(id, UsdLuxTokens->angle).Get<float>();
+        // TODO: implement physically correct conversion
+        float shadowSoftness = std::min(angle * (M_PI / 180.0) * M_PI, 1.0);
```
hshakula: Postponed until FIR-1579 resolved. For now, it implemented in such a way that shadows match Karma

PR #132, opened 2019-11-07
OpenVDB support

hshakula:
* #87 
* Houdini SOP (in-memory) openvdb volumes
* Varying in time volume fields

Currently, transform and material is not handled, yet.
To apply transformation correctly we need to apply it to both volume and base mesh including [voxels transformation](https://github.com/hshakula/RadeonProRenderUSD/blob/c03daa8652c801c1dba8b441b1364e09eae22bb8/pxr/imaging/plugin/hdRpr/rprApi.cpp#L759-L761). The cleanest way to achieve it is to create separate rpr::Volume class that will store the required information, i.e. use rpr::Volume instead of abstract RprApiObject.


bsavery:
So this includes @minghao-AMD s change then?

hshakula:
Yep

PR #139, opened 2019-11-12
Add "Pause/Resume Render" functionality

### LINKED ISSUE #74: Add "Pause/Resume Render" functionality to render delegate

hshakula:
* Move rendering to separate thread that can be paused/unpaused and stopped when needed.
    HdRprRenderBuffer now is simple proxy primitive that shared between rendering thread and main Hydra thread
* Use HdRenderParam to pass HdRprApi and HdRenderThread objects across primitives
    So instead of storing a weak pointer to HdRprApi in each primitive, HdRprApi queried from HdRenderParam
* Whenever the scene is going to be changed render thread must be stopped

Resolves #74

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -115,6 +115,7 @@ pxr_plugin(hdRpr
         rendererPlugin
         renderDelegate
         renderPass
+        renderThread
         rprApi
         mesh
         instancer
--- a/pxr/imaging/plugin/hdRpr/basisCurves.cpp
+++ b/pxr/imaging/plugin/hdRpr/basisCurves.cpp
@@ -1,14 +1,15 @@
 #include "basisCurves.h"
 #include "materialFactory.h"
 #include "material.h"
+#include "renderParam.h"
+#include "rprApi.h"
 
 PXR_NAMESPACE_OPEN_SCOPE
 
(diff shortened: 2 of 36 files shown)


hshakula:
Not ready to merge.
HdRenderThread has flaw implementation and can not be used as it is. I need to write my own implementation of it to fix all issues with it.

hshakula:
Ready to merge, I have added render thread implementation with appropriate fixes
* Pause waiting implemented without infinite loop which sleeps for some time to recheck pause condition. std::condition_variable used instead 
* Stop callback which called whenever a user asks thread to stop. In such a way, rprContextRender could be aborted to minimize render stopping time

hshakula:
1. In places where we store matrix to simply cache it and then pass to HdRprApi when needed, there is no point to cache double matrices because RPR API uses float matrices
2. HdRenderThread forces you to write ugly code to handle pause:
```cpp
while (renderThread->IsPauseRequested()) {
    if (renderThread->IsStopRequested()) {
        break;
    }
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
}
```
In my implementation is looks like this:
```cpp
std::unique_lock<std::mutex> lock(m_pauseWaitMutex);
while (m_pauseRender && !IsStopRequested()) {
    m_pauseWaitCV.wait(lock);
}
```
Also, HdRenderThread does not imply the existence of any stop callback that can be used to abort render when stop requested.

PR #140, opened 2019-11-12
Draft CPack package installer

hshakula:
Here is one of the possible ways to create hdRpr package to distribute it. Here are [examples](https://drive.google.com/open?id=1EgFY92R_xEjdP0mxem4DusyaA97clnHI) of created packages for windows and ubuntu. They can be created using command `cpack -C Release` in the build folder. Let's discuss


cmake/defaults/Version.cmake
```diff
@@ -22,6 +22,6 @@
 # language governing permissions and limitations under the Apache License.
 #
 # Versioning information
-set(USD_ARNOLD_MAJOR_VERSION "0")
-set(USD_ARNOLD_MINOR_VERSION "0")
-set(USD_ARNOLD_PATCH_VERSION "3")
+set(HD_RPR_MAJOR_VERSION "0")
```
bsavery: Versioning should be different for houdini build and RPR build you think?  Or should be 18.0.0 for Houdini 18 build?

pxr/imaging/plugin/hdRpr/package/readme.txt
```diff
@@ -0,0 +1 @@
+readme test
```
bsavery: ? Is this intended?

cmake/defaults/Version.cmake
hshakula: Potentially, hdRpr can be used in any software, so I don’t think that binding version to any particular soft like Houdini is a good idea. And having separate versioning seems like an additional headache

pxr/imaging/plugin/hdRpr/package/readme.txt
hshakula: No, I don't think so, that was for testing purposes, I will remove it or find usage for it tomorrow

bsavery:
This is ready to merge?

hshakula:
Yep

PR #143, opened 2019-11-14
Add "render device" and "render quality" controls on Houdini UI

### LINKED ISSUE #52: Houdini: add ability to change render device and render quality
### Description of Issue
The actual issue here is to found out the way how could we reemit a scene after render device change to setup scene for a newly created context

hshakula:
If there would no problems with RPR API when Hybrid plugin used, I would say that this is all.
This PR not ready to merge, because any Hybrid quality will crash now, due to numerous issues in RPR API, e.g. FIR-1589

This PR going to resolve #52


hshakula:
Ready for review

hshakula:
Now, when I already had the opportunity to work with switching between Full and other render qualities, I would say that two separate entries in the plugin list would be much more convenient than a current implementation with a manual restart requirement. Each time you want to switch from Full to Low you need to open render settings panel, switch render quality, open 'Persp' controls and select 'Restart Render'.

I think we should reconsider using two separate plugins, it could be much more convenient for the end-users. Let's ask them, somehow.

Also, there is such a thing as "[Material Binding Purpose](https://graphics.pixar.com/usd/docs/UsdShade-Material-Assignment.html#UsdShadeMaterialAssignment-Refinement2:MaterialPurpose)", currently, it always `full` in hdRpr, I think that for the hybrid plugin it would be more logical to set it to `preview`. And in such a way we would have hdRprFull and hdRprPreview plugins

PR #149, opened 2019-11-26
Update camera when viewport changed

hshakula:
If viewport size is changed perspective camera should be updated accordingly to a new aspect ratio

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -1320,7 +1320,8 @@ public:
     }
 
     void UpdateCamera() {
-        if ((m_dirtyFlags & ChangeTracker::DirtyCamera) == 0) {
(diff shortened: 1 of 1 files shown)


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1320,7 +1320,8 @@ class HdRprApiImpl {
     }
 
     void UpdateCamera() {
-        if ((m_dirtyFlags & ChangeTracker::DirtyCamera) == 0) {
+        if ((m_dirtyFlags & ChangeTracker::DirtyCamera) == 0 &&
```
bsavery: Should be this an || ?

hshakula: No, if both flags are not set we return from function

bsavery: Sorry, yes.  reverse logic.

PR #153, opened 2019-11-27
Update to RPR 1.34.4

hshakula:


cmake/modules/FindRpr.cmake
```diff
@@ -45,7 +45,6 @@ if(WIN32)
 
     set(RPR_BINARIES
         ${RPR_BIN_LOCATION}/RadeonProRender64.dll
-        ${RPR_BIN_LOCATION}/RprLoadStore64.dll
```
hshakula: Unused, no need to install

pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -471,7 +471,7 @@ class HdRprApiImpl {
             m_rprContext->GetHandle(), &curve
             , newPoints.size(), (float*)newPoints.data(), sizeof(GfVec3f)
             , newIndexes.size(), 1, (const rpr_uint*)newIndexes.data()
-            , &width, nullptr, segmentsPerCurve.data()), "Fail to create curve")) {
+            , &width, nullptr, segmentsPerCurve.data(), 0), "Fail to create curve")) {
```
bsavery: Does Hydra provide data for variable width curves?  If so in future we should support.

hshakula: As I can see, there are some widths used, but for some reason used only one width for the whole curve where width is average of all widths that Hydra provides us. I have no idea why it's done in such a way. Will create an issue for it

PR #168, opened 2019-12-02
Fix mesh displacement material setting

hshakula:
Displacement material relies on the mesh subdivision factor. Whenever the subdivision factor is 0 we need to unset displacement material and vice versa if the subdivision factor is greater than 0 we should set displacement material.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/materialFactory.cpp
+++ b/pxr/imaging/plugin/hdRpr/materialFactory.cpp
@@ -304,12 +304,22 @@ void RprMaterialFactory::DeleteMaterial(RprApiMaterial* material) {
 
 void RprMaterialFactory::AttachMaterialToShape(rpr_shape mesh, const RprApiMaterial* material) {
     if (material) {
-        rprShapeSetMaterial(mesh, material->rootMaterial);
-        if (material->displacementMaterial) {
-            rprShapeSetDisplacementMaterial(mesh, material->displacementMaterial);
+        RPR_ERROR_CHECK(rprShapeSetMaterial(mesh, material->rootMaterial), "Failed to set shape material");
+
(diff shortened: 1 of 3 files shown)


bsavery:
Just a question here.  Why do we *NEED* to unset displacement if subdivision is 0?  Would it make more sense to add subdivision factor automatically if displacement is set?

hshakula:
> Why do we NEED to unset displacement if subdivision is 0?

[RPR docs](https://radeon-pro.github.io/RadeonProRenderDocs/sdk/api/rprshapesetdisplacementmaterial.html) saying that subdivision is required for displacement to work. When I have been setting displacement to meshes without subdivision rprContextRender was not rendering anything without any errors. As soon as I unset displacement it would work as before.

> Would it make more sense to add subdivision factor automatically if displacement is set?

'Explicit is better than implicit' (c)
So I thought it would be better to ignore displacement on non-subdivision surfaces. And otherwise, if we would enable implicit subdivision, what subdivision level should we set?

bsavery:
Generally I agree in the explicit is better than implicit.  However in this case the end effect is no displacement when the user is expecting it.  So either way the behavior is unexpected for the user.  

My recommendation would be to implicitly add subdivision if displacement is set.

hshakula:
It makes sense for me, I agree. I will rewrite it in such a way that when the displacement is used without subdivision then it will force subdivision with refine level equal to 1. In such case, the user will see at least something, probably not perfect displacement but some low poly version of it, which should reveal the right direction to solve this issue - increase subdivision level

bsavery:
Agreed

hshakula:
Okay, I made it in such a way that it will try to set subdivision factor to 1 and only if it succeeds displacement material will be set

PR #170, opened 2019-12-03
Fix RIF image creation

hshakula:
After [RIF-#330](https://github.com/Radeon-Pro/RadeonProImageProcessing/pull/330) row_pitch value is in bytes, previously it was in pixels.
Also, automatic row_pitch calculation has been introduced, i.e. when row_pitch is 0 RIF will calculate it by himself

This PR should be merged when we move to RIF 1.4.3+


hshakula:
And we want to move to RIF 1.4.3 asap because it has fixes for AI denoiser resizing ([RIF-#338](https://github.com/Radeon-Pro/RadeonProImageProcessing/pull/338))

PR #171, opened 2019-12-03
Fix building of the plugin without OpenVDB support

hshakula:


pxr/imaging/plugin/hdRpr/houdini/CMakeLists.txt
```diff
@@ -1,3 +1,8 @@
+if(RPR_ENABLE_OPENVDB_SUPPORT)
```
bsavery: IMO RPR_ENABLE_OPENVDB_SUPPORT should always be true if Houdini build no? 

Only usdview should have optional openvdb

hshakula: I have enabled OpenVDB by default when building Houdini plugin in abed8b0
But I left possibility for the user to disable it

bsavery: Ok, not sure why they'd want to but that makes sense.

PR #172, opened 2019-12-03
chore: Allow config specification for package generation

hshakula:


pxr/imaging/plugin/hdRpr/package/generatePackage.py
```diff
@@ -50,15 +50,19 @@ def get_package_path(build_dir):
...
 if package_path:
     package_basename = os.path.basename(package_path)
 
     output_package_path = args.output_package_path
     if not output_package_path:
-        package_name = os.path.splitext(package_basename)[0]
+        package_name = package_basename
+        package_ext = '.tar.gz'
+        if package_basename.endswith(package_ext):
+            package_name = package_basename[:-len(package_ext)]
```
hshakula: Remove '.tar.gz' from package basename instead of removing just '.gz'

PR #174, opened 2019-12-04
Add ability to dump USD errors to a specified file

hshakula:
This very useful feature that allows getting important info from hdRpr users in case of any issues,
especially in Houdini where all USD errors go to console which could be inaccessible in case of crashes.
Currently, there is no such embedded functionality in USD


bsavery:
So the user would set HD_RPR_ERROR_OUTPUT_FILE = some_file.txt ?

hshakula:
Yeah, stdout and stderr is viable options as well

hshakula:
Yeah, stdout and stderr is viable options as well

PR #178, opened 2019-12-06
Remove undesirable implicit subdivision on non-displaced meshes

hshakula:
This bug comes from #168. Due to logical error, the subdivision was enabled even for meshes without displacement material


hshakula:
It would be nice to rebase(fixup) this commit into 8d6b3c7 because on its own this commit do no have any particular interest from a repository history standpoint

hshakula:
But thanks to this bug I have found out another bug in RPR API - [FIR-1601](https://amdrender.atlassian.net/browse/FIR-1601)

PR #179, opened 2019-12-09
Add support for camera parameters that HdCamera missing

### LINKED ISSUE #164: Camera Focal length missmatch Karma/RPR
I have a weird issue with camera focal length, switching between Karma and RPR they do not match. (screenshots and scene file attached)

lopnet_Karma
lopnet_RPR
stage_Karma
stage_RPR
[RPR_Test_01.hiplc.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/3907147/RPR_Test_01.hiplc.zip)

hshakula:
Default Hydra Camera (**HdCamera**) does not support parameters defined in **UsdGeomCamera** such as:
* Aperture Size
* Focal Length
* FStop
* Focus Distance
* Shutter Open
* Shutter Close
* Clipping Range

To process these parameters custom **HdRprCamera** class that inherits **HdCamera** is introduced. This class simply adds syncing of missing parameters so that they could be used by the render engine.

Resolves #164

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -8,6 +8,7 @@
 #include "rifcpp/rifError.h"
 
 #include "config.h"
+#include "camera.h"
 #include "imageCache.h"
 #include "material.h"
 #include "materialFactory.h"
@@ -22,6 +23,8 @@
 #include "pxr/base/plug/plugin.h"
 #include "pxr/base/plug/thisPlugin.h"
 #include "pxr/imaging/glf/uvTextureData.h"
+#include "pxr/usd/usdRender/tokens.h"
+#include "pxr/usd/usdGeom/tokens.h"
 
 #include <RadeonProRender.h>
 #include <RadeonProRender_Baikal.h>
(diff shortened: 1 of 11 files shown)


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1070,40 +1087,123 @@ class HdRprApiImpl {
...
+                m_cameraProjectionMatrix[1][1] = 2.0f / apertureSize[1];
+                m_cameraProjectionMatrix[2][2] = -2.0f / (zFar - zNear);
+                m_cameraProjectionMatrix[3][2] = -(zFar + zNear) / (zFar - zNear);
+            } else {
+                if (projectionType != UsdGeomTokens->perspective) {
+                    TF_CODING_ERROR("Unexpected projection type: %s. Fallback value: perspective", projectionType.GetText());
+                    projectionType = UsdGeomTokens->perspective;
+                }
+
+                float range = (apertureSize[1] * 0.5f) / focalLength * zNear;
```
bsavery: Do you think these calculations should be in rprCamera.cpp?

hshakula: Yeah, it should be there

hshakula: I will refactor it later today

hshakula: These calculations could not be moved to HdRprCamera because they rely on the viewport size which currently available only in HdRprApi and HdRprRenderPass. Right now, HdRprCamera is a simple sync&store entity

bsavery: Ah.  That makes sense.

PR #207, opened 2019-12-18
Handle image gamma for images loaded using GlfUVTexture

hshakula:


pxr/imaging/plugin/hdRpr/materialFactory.cpp
```diff
@@ -258,15 +258,6 @@ RprApiMaterial* RprMaterialFactory::CreateMaterial(EMaterialType type, const Mat
...
 
-        // SIGGRAPH HACK: Fix for models from Apple AR quick look gallery.
-        //   Of all the available ways to load images, none of them gives the opportunity to
-        //   get the gamma of the image and does not convert the image to linear space
-        if ((paramId == RPR_MATERIAL_INPUT_UBER_DIFFUSE_COLOR ||
-             paramId == RPR_MATERIAL_INPUT_UBER_REFLECTION_COLOR) &&
-            ArGetResolver().GetExtension(matTex.path) == "png") {
-            rprImageSetGamma(material->materialImages.back()->GetHandle(), 2.2f);
-        }
-
```
hshakula: There is no need for this hack anymore because we have only two ways to load image from file (GlfUvTextureData and rprContextCreateImageFromFile) and both of them handle image gamma

PR #211, opened 2019-12-19
Adding release notes.

bsavery:


CHANGELOG.MD
```diff
@@ -0,0 +1,43 @@
+# Change Log
+## 1.1 Release
+	- Supports Houdini 18.0 and USD 19.11
+	- Uses RPR SDK 1.34.5
+	- Uses RIF 1.4.3
```
hshakula: Let's move it to README.md

I was just thinking about it today. Why don't we update RPR and RIF version in README? There is a special column for that. It would be very convenient to track it in such a way so that when you are going through commits trying to find when exactly things get broken. Right now you have to lookup previous commits finding the commit that looks like "Update to RIF (version)" to get the version of SDKs that was used to build current commit. Of course, the submodule of RadeonProRenderThirdPartyComponents would do it much better, but until we can use it, let's use README to track the current version of dependencies.

CHANGELOG.MD
```diff
@@ -0,0 +1,43 @@
...
+### New features
+	- Added Centos 7 build
+	- Added RPR Render Settings node for use in Houdini in addition to global per project settings.
```
hshakula: Actually, all render settings that were available globally in "Display Options" moved to native LOP node "Render Settings" into its own tab "RPR", so there are no render settings in `global per project settings`

CHANGELOG.MD
```diff
@@ -0,0 +1,43 @@
...
+	- Image Gamma supported for GL textures
+	- Add support for alpha compositing render output.  This allows overlaying on the Houdini background and viewport gizmos
+	- Support for UsdTransform2D nodes in shading (Only single level, not nested)
+	- Per object subdivision.  Also added a HDA RPR Mesh Properties Node for specifying subdivision level in Houdini 
+	- Support of doublesided meshes
+	- Support of HdDisplayStyle::displacementEnabled attributes
+	- Camera FStop, Focal Length, Aperature, Focus Distance and Clipping Range is supported on cameras
```
hshakula: Typo in "aperture"

CHANGELOG.MD
```diff
@@ -0,0 +1,43 @@
...
+	- Support for tapered width curves
```
hshakula: only linear curves right now. I will add support for bezier soon

CHANGELOG.MD
```diff
@@ -0,0 +1,43 @@
...
+	- Add debugging environment variable.  Set HD_RPR_ERROR_OUTPUT_FILE=some_file.txt to dump errors when rendering with hdRPR.
```
hshakula: I would specify that it works only for USD errors and warnings - all errors that come from TF_RUNTIME_ERROR, TF_CODING_ERROR, TF_WARN, etc. Why? Because they are typically hidden by Houdini and not always printed into "Houdini Console"

CHANGELOG.MD
```diff
@@ -0,0 +1,43 @@
...
+	- 
+
+### Fixes
+	- Fixed finding Hybrid.so on Ubuntu
+	- Fixed teardown of Denoiser filter when disabled
+	- Fixed material attachment to instances in hybrid modes
+	- Fixed AOV creation in CPU rendering
```
hshakula: This is the fixes of bugs that popped after my latest changes, this was not in 1.0 release. So I don't think it should be here.

bsavery:
Updated

CHANGELOG.MD
hshakula: This one is multiline

CHANGELOG.MD
```diff
@@ -0,0 +1,36 @@
...
+	- Support of HdDisplayStyle::displacementEnabled attributes
+	- Camera FStop, Focal Length, Aperture, Focus Distance and Clipping Range is supported on cameras
+	- Support for tapered width curves (linear curves only)
+	- Add debugging environment variable.  Set HD_RPR_ERROR_OUTPUT_FILE=some_file.txt to dump USD errors when rendering with hdRPR in Houdini.
+	
+
+### Fixes
+	- Fixed finding Hybrid.so on Ubuntu
+	- Fixed teardown of Denoiser filter when disabled
+	- Fixed material attachment to instances in hybrid modes
```
hshakula: I've pointed to 3 lines actually, not to only "Fixed AOV creation in CPU rendering" line
Github highlighting is confusing here

bsavery: got it.

hshakula:
Let's wait for CIS first?

PR #221, opened 2020-01-10
Fix render quality switching

hshakula:
* Correctly clear AOVs when switching between hybrid qualities
* Do not clear AOVs when switching into the state that requires a restart


bsavery:
It looks like there are some conflicts wiht the medium quality option?  Not sure what is intended here.

hshakula:
Fixed

PR #223, opened 2020-01-10
Remove brutforce search of the working one plugin when context with requested plugin is can't be created

hshakula:
In back old days this behavior was pretty convenient, but right now it confusing.
For example, you have selected High quality, i.e. use Hybrid plugin, but in case Hybrid plugin can't be loaded for some reason Tahoe would be loaded automatically and there would be no sign of problem, of course except logged error message which Houdini tends to hide


bsavery:
So just to be clear.  I think we should grey our or remove the quality options if only tahoe is available...

hshakula:
Sure, it would be nice to have such a possibility but neither usdview nor Houdini allows us to modify its UI dynamically:
* Houdini loads UI from "template" file and creates UI from it
* usdview generates UI from the set of RenderSettings returned by render delegate (`HdRenderDelegate::GetRenderSettingDescriptors`)

PR #237, opened 2020-02-21
Fix adaptive sampling

hshakula:
* Tahoe requires that `AOV_VARIANCE` is created for adaptive sampling to work
* Set `RPR_CONTEXT_FRAMECOUNT` parameter to the current sample count (should not `RPR_CONTEXT_FRAMECOUNT` be named as `RPR_CONTEXT_SAMPLECOUNT`?)


bsavery:
I agree, the name of RPR_CONTEXT_FRAMECOUNT is confusing.  Probably should be RPR_CONTEXT_ITERATIONCOUNT since you can have multiple samples per iteration.

PR #239, opened 2020-02-21
Add "restart required" popup

hshakula:
Implement long-living Fianna's request - popup window when you have to restart renderer.
On windows, I use `MessageBox` from the Windows Forms library.
On macOS AppKit's `NSAlert`.
Linux is the hard one, there are no such system GUI APIs on Linux. The easiest way I've found is to use `xmessage`.


bsavery:
A stupid question of course is that we can't automatically restart if device changes?

hshakula:
Unfortunately, we can't.

btw, that's how these  warning messages looks:
macosWarning
winWarning
linuxWarning

PR #240, opened 2020-02-24
Allow using hdRpr plugin on Linux from non-standard paths

hshakula:
## Motivation

Right now, if the user has Houdini installed under the path that is not /opt/hfs18.0 or not the one we built against, hdRpr.so will fail to load.

##### master
```
- RUNPATH $ORIGIN/../../lib:/home/hshakula/dev/builds/RPR_SDK_1.34.7/binUbuntu18:/home/hshakula/dev/builds/RIF_SDK_1.4.4/Ubuntu18/Bin/Release/x64:/opt/hfs18.0/dsolib:/opt/hfs18.0.387/dsolib
```

The issue is here `/opt/hfs18.0/dsolib:/opt/hfs18.0.387/dsolib`, in absolute paths which bounds us to the particular naming of Houdini install root.

`/home/hshakula/dev/builds/RPR_SDK_1.34.7/binUbuntu18:/home/hshakula/dev/builds/RIF_SDK_1.4.4/Ubuntu18/Bin/Release/x64` is just a trash that should not be in rpaths.

##### fix-linux-rpath
```
- RUNPATH $ORIGIN/..
```

Houdini's directory structure looks like this:
```
Root
 - dsolib
  - usd_plugins
   - hdRpr
   - hdRpr.so
   - ...
  - ...
 -...
```
So new rpath will point to dsolib and it's not bound to the Root name at all.

##### master
```
-RUNPATH $ORIGIN/../../lib:/home/hshakula/dev/builds/RPR_SDK_1.34.7/binUbuntu18:/home/hshakula/dev/builds/RIF_SDK_1.4.4/Ubuntu18/Bin/Release/x64:/home/hshakula/dev/USD_build/lib
```
##### fix-linux-rpath
```
- RUNPATH $ORIGIN/../../lib
```

Everything the same, just without the trash.

### macOS

macOS, in particular, is an interesting case because it has SIP which disallows non-absolute rpaths.
So in the case of macOS, hdRpr pre-built plugin is always bound to a particular Houdini version.
On macOS effective change is the removal of the trash.

##### master
```
- path /Applications/Houdini/Houdini18.0.350/Frameworks/Houdini.framework/Versions/18.0/lib (offset 12)
- path /Users/hshakula/dev/builds/RPR_SDK_1.34.7/binMacOS (offset 12)
- path /Users/hshakula/dev/builds/RIF_SDK_1.4.4/OSX/Bin/Release/x64 (offset 12)
- path /Applications/Houdini/Houdini18.0.350/Frameworks/Houdini.framework/Versions/18.0/Resources/../Libraries (offset 12)
- path /Applications/Houdini/Houdini18.0.350/Frameworks/Houdini.framework/Versions/18.0/Libraries (offset 12)
- path /Applications/Houdini/Houdini18.0.350/Frameworks (offset 12)
```
##### fix-linux-rpath
```
- path /Applications/Houdini/Houdini18.0.350/Frameworks/Houdini.framework/Versions/18.0/Libraries (offset 12)
```
##### master
```
- path /Users/hshakula/dev/USD_build/lib (offset 12)
- path /Users/hshakula/dev/builds/RPR_SDK_1.34.7/binMacOS (offset 12)
- path /Users/hshakula/dev/builds/RIF_SDK_1.4.4/OSX/Bin/Release/x64 (offset 12)
```
##### fix-linux-rpath
```
- path /Users/hshakula/dev/USD_build/lib (offset 12)
```

## Summary

Fix rpath on Linux and tweak it on macOS.
Regarding macOS absolute rpath, I hope there is some way to work around this issue but it's not the original issue I was going to fix with this PR.
I will add a corresponding entry in my TODO-list to check what can be done here.


bsavery:
Regarding macOS.  I'm not sure I now what you mean about needing an absolute path.  You can have an $RPATH in paths no?

hshakula:
When I mentioned the "macOS issue" I was referring to this part of the [cmake code](https://github.com/PixarAnimationStudios/USD/blob/master/cmake/macros/Private.cmake#L628). You can see that Pixar absolutizes rpaths.
Exactly this line binds our hdRpr build on macOS to a particular Houdini version, so before removing it, I need to check how it works in Houdini what I decided to do as part of a separate task.

PR #255, opened 2020-03-11
Fix state initialization

hshakula:


bsavery:
Can you explain why this change is needed?  What problem is it fixing?

hshakula:
Sure, have a look at this [fast-return-if](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/fb4ec170b79645391158023b7cbb488072cf3e70/pxr/imaging/plugin/hdRpr/rprApi.cpp#L126) - it's fully based on the assumption that `m_state != kStateUninitialized` is true only when we already tried to initialize RPR.

So, I've faced an issue when I have a few threads and one of them has locked mutex and set `m_state` to `kStateRender` before even doing the initialization. Just after this, other threads encounter "fast-return-if" and, obviously, they do not wait (on locked mutex) for initialization when they should. In such a way some calls would silently do nothing because we have similar "fast-return-if" (checking for `m_rprContext`) in a lot of other places (e.g. [this](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/fb4ec170b79645391158023b7cbb488072cf3e70/pxr/imaging/plugin/hdRpr/rprApi.cpp#L156)).

While I wrote the text above, I understood that this PR will change behavior when RPR fails to initialize. If previously it would fail one time and never try again until render delegate would be recreated, then now it will try to it again and again. What is the behavior we would like to have?

bsavery:
> While I wrote the text above, I understood that this PR will change behavior when RPR fails to initialize. If previously it would fail one time and never try again until render delegate would be recreated, then now it will try to it again and again. What is the behavior we would like to have?

This is a good question.  One could assume that if it fails to initialize once, it will fail again and again.

PR #258, opened 2020-03-13
Refactor camera projection matrix calculation

hshakula:
Projection matrix stored in HdCamera, whom delegate is Houdini's scene delegate, has been fixed and therefore can be used directly


bsavery:
Does this require a specific version of houdini?

hshakula:
I’ve tested on the latest production release (18.0.391)

bsavery:
Ok let's just make sure we note needed version for users when we put a release up.

hshakula:
Starting from this https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/244 , Houdini version is in the package name

PR #261, opened 2020-03-18
Add support for visibility flags

hshakula:
Hydra implementation side:
* UsdPrim's visibility is still accounted
    - If primitive is invisible it overrides visibility flags, otherwise visibility flags in priority
* Visibility flags parsed from `rpr:visibilityMask` primvar
* Visibility mask primvar supported by meshes, curves and geometry lights (sphere, rect, disk, cyl)
* Remove `HDRPR_VISIBLE_LIGHTS` as discussed here https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/238

Houdini UI side:
* In Houdini, it's possible to add renderer custom settings for lights the same as render settings - via .ds file but with the corresponding name. In our case `HdRprPlugin_Light.ds`.
    By using it we expose visibility mask under the RPR tab in the "Light" LOP node.
* Also, Houdini has "Geometry Settings" LOP node that does the same with `HdRprPlugin_Geometry.ds`, so we expose visibility mask for meshes under the RPR tab in the "Geometry Settings" node.
* Instead of having separate "RPR Mesh Properties" node in LOP, all its parameters have been moved to the "Geometry Settings" node.


pxr/imaging/plugin/hdRpr/python/commonSettings.py
```diff
@@ -0,0 +1,32 @@
...
+    'defaultValue': '*',
+    'hints': [
+        ('*', 'Visible to all'),
+        ('primary', 'Visible only to primary rays'),
+        ('primary,shadow', 'Visible only to primary and shadow rays'),
+        ('-primary', 'Invisible to primary rays'),
+        ('-primary,-shadow,-light', 'Invisible emissive light'),
+        ('', 'Invisible'),
+    ],
+    'help': 'The visibility mask is a comma-separated list of inclusive or exclusive ray visibility flags.\\n' \
```
bsavery: I think my only question here is only that all are combined to one setting.  Would it make more sense to have different boolean settings for each visibility flag?

hshakula: I thought about implementing it in such a way but a big number of boolean parameters are always very clunky. Imagine that you want to enable just one type of rays, you will need to uncheck all other parameters. It seemed to me like too many clicks.
So I decided to implement it in a very much like in Karma.
If you prefer it to be on the GUI as booleans parameters I can easily change it

bsavery: I think I do.  Mainly because of what we do in other plugins and Renderman, arnold plugins do it this way. At least renderman does.

bsavery: Not to be one to follow the leader but for the simple user this is easier.

hshakula: ok, I will change it

PR #268, opened 2020-03-31
Improve HdVolume handling

hshakula:
* Add field parameters:
    - normalize: whether fieldValue should be normalized
    - scale: scale to be applied to value before the lookup table.
    - gain: gain to be applied to LUT values.
    - bias: bias to be applied to LUT values.
    - ramp: value lookup table.
    - blackbodyMode: parameters that control LUT for temperature grid
* Add volume parameters:
    - scatteringColor
    - transmissionColor
    - emissionColor
    - anisotropy
    - multipleScattering
* Add parsing of field parameters from .vdb file metadata
* HdVolume now reacts to changes in HdField it uses
* HdVolume now reads a correct field
* Avoid allocating memory for the same indices arrays when duplicating topology (by using VtArray instead of std::vector)


bsavery:
Please note used packages in license.MD

bsavery:
Also any before and after renders?

LICENSE.md
```diff
@@ -199,4 +199,34 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE  IS PROVIDED "AS  IS", WITHOUT WARRANTY  OF ANY KIND,  EXPRESS OR
+IMPLIED,  INCLUDING BUT  NOT  LIMITED TO  THE  WARRANTIES OF  MERCHANTABILITY,
+FITNESS FOR  A PARTICULAR PURPOSE AND  NONINFRINGEMENT. IN NO EVENT  SHALL THE
+AUTHORS  OR COPYRIGHT  HOLDERS  BE  LIABLE FOR  ANY  CLAIM,  DAMAGES OR  OTHER
+LIABILITY, WHETHER IN AN ACTION OF  CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE  OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
hshakula: Have I added the correct dependency pointer?

hshakula:
You will see before and after changes only on the scenes with such things:
* Custom fieldName - so basically when this PR applied, hdRpr will fetch correct fieldName and correspondingly render the correct image
* .vdb file with Houdini-style metadata and authored emission color ramp. Attaching some images for this one:
    - After  after
    - Before (in direct comparison, pure black due to the absence of lights on the original scene) before
    - Before (with light), as you can see volume is not emitting because temperature grid values require special temperature lookup table that is distributed within .vdb file
before_with_light
    - Houdini GL houdiniGL

hdRpr still does not match ideally to Houdini's renders. Mainly, this difference is caused by the fact that Houdini uses two distinct fields to control emissiveness - temperature grid and glowness grid (how bright to glow each voxel). The two values are multiplied together to get the emission at a voxel.

RPR does not allow any math logic with rpr_grids. Please note that we can't simply multiply two fields as preprocessing step because we need to multiply values after field lookup.

LICENSE.md
bsavery: Yes.

PR #275, opened 2020-04-13
Add ability to export .rpr file with the current scene

hshakula:
* Add C-API interface to enqueue .rpr file export - `int HdRprExportRprSceneOnNextRender(const char* exportPath)` - on the next Render call, scene will be exported to the provided `exportPath`
* Add Houdini "RPR Export .rpr" LOP node with file path parameter. Here is the python code that is run when this node is added to the Houdini node graph:

```python
from ctypes import cdll, c_int, c_char_p
from pxr.Plug import Registry
import os

def RprExportRprSceneOnNextRender(path):
    rprPluginType = Registry.FindTypeByName('HdRprPlugin')
    plugin = Registry().GetPluginForType(rprPluginType)
    if plugin and plugin.path:
        lib = cdll.LoadLibrary(plugin.path)
        if lib:
            lib.HdRprExportRprSceneOnNextRender.restype = c_int
            err = lib.HdRprExportRprSceneOnNextRender(c_char_p(path))
            if err == 0:
                print('"{}" export successfully enqueued.'.format(path))
                return True
            elif err == -1:
                print('Failed to export: hdRpr is not enabled');
            else:
                print('Failed to export: HdRprExportRprSceneOnNextRender ret={}'.format(err));
        else:
            print('Failed to export: can not load RPR plugin lib: {}'.format(plugin.path))
    else:
        print('Failed to export: can not find plugin')
    return False

def Export(node):
    exportPath = node.evalParm('exportPath')
    if not exportPath:
        return
    
    filename, ext = os.path.splitext(os.path.basename(exportPath))
    if ext != '.rpr':
        print('Failed to export: invalid path - ".rpr" extension required')
        return

    if not RprExportRprSceneOnNextRender(exportPath):
        return
    
    # Force render
    stage = node.editableStage()
    cube = stage.DefinePrim('/rprExportImplicitPrimitive/' + filename, 'Cube')
    cubeSizeAttr = cube.GetAttribute('size')
    size = cubeSizeAttr.Set(0.0)

node = hou.pwd()
if node.isDisplayFlagSet():
    Export(node)
```


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1430,6 +1431,19 @@ class HdRprApiImpl {
...
         }
 
+        if (!m_rprSceneExportPath.empty()) {
+            std::string exportPath;
+            {
+                std::lock_guard<std::mutex> lock(m_rprSceneExportPathMutex);
+                std::swap(m_rprSceneExportPath, exportPath);
+            }
+            if (!exportPath.empty()) {
+                auto rprContextHandle = rpr::GetRprObject(m_rprContext.get());
```
hshakula, line 1441: I've marked this PR as "draft" because of this line. This line will compile only when https://github.com/Radeon-Pro/RadeonProRenderAPI/pull/68 is merged

hshakula:
Can be merged after https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/280

PR #276, opened 2020-04-17
Added initial interop support

s1lentssh:


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -137,6 +139,22 @@ class HdRprApiImpl {
         //InitIfNeeded();
     }
 
+	rpr::FrameBuffer* GetColorFramebuffer() {
+		auto aovIter = m_boundAovs.find(TfToken("color"));
```
hshakula, line 143: Use existing `color` token instead. `HdAovTokens->color` for example

pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1373,6 +1391,17 @@ class HdRprApiImpl {
...
+
+		// Made temporarily just to have working interop
+		if (m_rprContextMetadata.pluginType == rpr::kPluginHybrid) {
+			std::cout << "[Plugin] Render\n";
+			auto status = m_rprContext->Render();
+			rprContextFlushFrameBuffers_func rprContextFlushFrameBuffers;
+			rprContextGetFunctionPtr(m_rprContext->Handle(), RPR_CONTEXT_FLUSH_FRAMEBUFFERS_FUNC_NAME, (void**)(&rprContextFlushFrameBuffers));
+			rprContextFlushFrameBuffers(m_rprContext->Handle());
+			return;
+		}
```
hshakula, line 1403: This should be called after each `Render` or just once? What's the logic behind this flushing?

pxr/imaging/plugin/hdRpr/renderDelegate.cpp
```diff
@@ -401,6 +400,17 @@ bool HdRprDelegate::Resume() {
     return true;
 }
 
+void HdRprDelegate::SetDrivers(HdDriverVector const& drivers) {
+	for (HdDriver* hdDriver : drivers) {
+		if (hdDriver->name == TfToken("RPR") && hdDriver->driver.IsHolding<VtDictionary>()) {
+			VtDictionary dictionary = hdDriver->driver.UncheckedGet<VtDictionary>();
```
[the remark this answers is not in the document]
s1lentssh: VtDictionary operator[] has no const overload, maybe it isn't a big performance issue here?

pxr/imaging/plugin/hdRpr/rpr/contextHelpers.cpp
```diff
@@ -21,6 +21,10 @@ limitations under the License.
 #include "pxr/base/tf/envSetting.h"
 
 #include <RadeonProRender.hpp>
+#include <RadeonProRender_VK.h>
+#include <RadeonProRender_Baikal.h>
+#include <vulkan/vulkan.h>
+#include <iostream>
```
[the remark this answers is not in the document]
s1lentssh: It was for debug purposes, removed

pxr/imaging/plugin/hdRpr/rprApi.cpp
s1lentssh: Done

pxr/imaging/plugin/hdRpr/rprApi.cpp
s1lentssh: Still figuring out. Render should be in loop till enough render passes reached. Then flush function should be called. Flush  make semaphores signalled and tells us we can render next frame.

pxr/imaging/plugin/hdRpr/renderDelegate.cpp
```diff
@@ -401,6 +400,17 @@ bool HdRprDelegate::Resume() {
     return true;
 }
 
+void HdRprDelegate::SetDrivers(HdDriverVector const& drivers) {
+	for (HdDriver* hdDriver : drivers) {
+		if (hdDriver->name == TfToken("RPR") && hdDriver->driver.IsHolding<VtDictionary>()) {
```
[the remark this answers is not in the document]
s1lentssh: Done

s1lentssh:
> * If I understand correctly, with these changes, hdRpr will not work correctly (`contextProperties` with null `RPR_CONTEXT_CREATEPROP_VK_INTEROP_INFO`, `yflip == 1`) with the Hybrid plugin if the owner of a render delegate does not provide correct HdDriver, isn't it?

Yes, i think if we would like to support interop mode for hybrid, no-interop mode should be unsupported, but it might require discussion.

PR #277, opened 2020-04-24
Added interop present synchronization

s1lentssh:
Added synchronization that prevents flushing two frames in series without presenting after each flush.
Added blocking wait for present on plugin side.

Condition variable is used to prevent this issue:
[Plugin] Render_Frame_1 & Flush_Frame_1
[Plugin] Render_Frame_2 & Flush_Frame_2
[Client] Present frame
Hybrid correct usage prohibit flushing next frame before previous was presented
Render thread would wait on next flush till previous frame would be presented, example:
[Plugin] Render_Frame_1 & Flush_Frame_1
[Plugin] Render_Frame_2 & [Wait for present] <- Here frame wasn't presented yet
[Client] Present Frame_1
[Plugin] [Wake up] Flush Frame_2, continue work


pxr/imaging/plugin/hdRpr/renderBuffer.cpp
```diff
@@ -124,6 +130,7 @@ VtValue HdRprRenderBuffer::GetResource(bool multiSampled) const {
         VtDictionary dictionary;
         dictionary["is_hybrid"] = isHybrid;
         dictionary["framebuffer"] = color;
+        dictionary["mutex_ptr"] = &color->GetContext().GetMutex();
```
hshakula, line 133: As far as I understood, you are using this mutex for two purposes:
1. Protect RPR API access [here](https://github.com/Radeon-Pro/RPRViewer/blob/99d9c855220bdea896bebc3728d56025d2c49e6e/RPRViewer/src/hdx/rprInteropPresenter.cpp#L58)
2. Protect access to `_presentedCondition`

This mutex will work for the second purpose but it will not work for the first one. This mutex does not actually protect RPR API calls because hdRpr uses another mutex for this purpose. hdRpr uses [this](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/rprApi.cpp#L79) mutex.

PR #281, opened 2020-05-14
Add RPR and RIF as submodules

hshakula:


bsavery:
I think the README should have instrcutions for `git submodule init `

PR #282, opened 2020-05-15
BasisCurves improvements

hshakula:
### New

* Support for curves of bezier basis
* Primitive support for catmullRom and spline bases by interpreting them as linear
* Support varying interpolation for widths
* Optimized a bit memory usage for linear curves - there is no need to [allocate memory](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/v1.2.2/pxr/imaging/plugin/hdRpr/basisCurves.cpp#L100) for dummy indices when there are no authored ones 

### Open Questions

* Should we consider converting (in the plugin) catmullRom and spline bases to bezier or the current behavior is sufficient?

### DIFF
--- a/pxr/imaging/plugin/hdRpr/basisCurves.cpp
+++ b/pxr/imaging/plugin/hdRpr/basisCurves.cpp
@@ -77,30 +77,7 @@ void HdRprBasisCurves::Sync(HdSceneDelegate* sceneDelegate,
         m_topology = sceneDelegate->GetBasisCurvesTopology(id);
         m_indices = VtIntArray();
         if (m_topology.HasIndices()) {
-            if (m_topology.GetCurveWrap() == HdTokens->nonperiodic || // GL_LINE_STRIP
-                m_topology.GetCurveWrap() == HdTokens->periodic) { // GL_LINE_LOOP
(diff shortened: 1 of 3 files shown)


hshakula:
Here's an example of bspline curves that are interpreted as linear for RPR

bspline_as_linear

Render from Karma

bspline_karma

Curves asset is a carpet from [Esper Room by Multiverse](https://j-cube.jp/solutions/multiverse/assets/)

bsavery:
How do Catmull Rom curves work now?
IMO RPR itself should do the conversion from segments to curved scheme

hshakula:
>How do Catmull Rom curves work now?

Same as bspline - data interpreted as linear

> IMO RPR itself should do the conversion from segments to curved scheme

Do you have examples?

bsavery:
Meaning you should load a bunch of points into RPR and tell it how to interpolate the curves, linear, etc https://rmanwiki.pixar.com/display/REN23/Curves

hshakula:
Well, that would be ideal, but for now, RPR curves can only be of bezier type. So even linear curves are not completely correct.

Correct me if I'm wrong

bsavery:
No you're correct.  I just am saying we should propose that to RPR SDK.

PR #283, opened 2020-05-18
Embed tone mapping filter support

hshakula:
While working with indoor scenes (like the attic and the esper room) I felt the need to apply tone mapping. So as RIF already has a bunch of tone mapping filter I tested a few of them and selected the most versatile one - [Linear Photographic Tone Mapping Filter](https://radeon-pro.github.io/RadeonProRenderDocs/rif/filters/linear_photographic_tone_mapping.html).

It's possible to add support for more tone mapping filters but I don't feel such need right now

### DIFF
--- a/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
+++ b/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
@@ -198,6 +198,61 @@ render_setting_categories = [
         ]
     },
     {
+        'name': 'Tonemapping',
+        'settings': [
+            {
+                'name': 'enableTonemap',
+                'ui_name': 'Enable Tone Mapping',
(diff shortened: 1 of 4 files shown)


bsavery:
What reason made you decide this vs the ACES filmic one?  The latter is often used as well.

hshakula:
When I first saw the list of tone mapping filter, I also thought that I would stick to ACES filter but when I tried it and, especially compared it to Linear Photographic, I understood that it's not an option because of how much better I can control the output image with almost the same set of options. The second reason is more subjective - I just liked the output more with this filter, It coincided more with the output which I expect.

As I can see RIF's ACES filter is just fitted curve from here - https://knarkowicz.wordpress.com/2016/01/06/aces-filmic-tone-mapping-curve/
But, as the author states, this curve is not the best choice for realistic rendering

PR #289, opened 2020-06-02
Add render mode render setting

hshakula:
I'm not exposing "Direct Illumination No Shadows" render mode because it does not work in core. See [FIR-1584](https://amdrender.atlassian.net/browse/FIR-1584)

Global Illumination

globalIllumination

Direct Illumination

directIllumination

Wireframe

wireframe

Ambient Occlusion

ao

Diffuse

diffuse

### DIFF
--- a/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
+++ b/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
@@ -16,6 +16,19 @@ import platform
 
 from houdiniDsGenerator import generate_houdini_ds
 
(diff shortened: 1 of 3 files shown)


bsavery:
Shouldn't diffuse be colored?  Maybe it's just diffuse amount.  So diffuse would diffuse * albedo color

hshakula:
Well, I'm not sure how it supposed to look but checking related JIRA tickets shows that it's expected output:
https://amdrender.atlassian.net/browse/FIR-815
https://amdrender.atlassian.net/browse/FIR-1421

Probably, it would be better to ask Takahiro

PR #292, opened 2020-06-02
Fix dome light visibility

hshakula:
Due to how is environment light creation/deletion implemented in RPR API we must remove default environment light before creating a new one

### DIFF
--- a/pxr/imaging/plugin/hdRpr/distantLight.cpp
+++ b/pxr/imaging/plugin/hdRpr/distantLight.cpp
@@ -61,7 +61,6 @@ void HdRprDistantLight::Sync(HdSceneDelegate* sceneDelegate,
                 *dirtyBits = HdLight::Clean;
                 return;
             }
-            rprRenderParam->AddLight();
         }
 
         float angle = sceneDelegate->GetLightParamValue(id, UsdLuxTokens->angle).Get<float>();
@@ -88,8 +87,6 @@ void HdRprDistantLight::Finalize(HdRenderParam* renderParam) {
         auto rprRenderParam = static_cast<HdRprRenderParam*>(renderParam);
         rprRenderParam->AcquireRprApiForEdit()->Release(m_rprLight);
         m_rprLight = nullptr;
-
-        rprRenderParam->RemoveLight();
     }
 
     HdSprim::Finalize(renderParam);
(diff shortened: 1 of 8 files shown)


pxr/imaging/plugin/hdRpr/light.cpp
```diff
@@ -385,43 +390,36 @@ void HdRprLight::Sync(HdSceneDelegate* sceneDelegate,
...
                 if (emissionColorIsDirty) { rprApi->SetLightColor(light, emissionColor); }
-                return true;
             }
         };
 
-        if (BOOST_NS::apply_visitor(LightParameterSetter{rprApi, emissionColor, isEmissionColorDirty}, m_light) && !m_created) {
-            m_created = true;
-            rprRenderParam->AddLight();
-        }
+        BOOST_NS::apply_visitor(LightParameterSetter{rprApi, emissionColor, isEmissionColorDirty}, m_light);
```
bsavery, line 422: Anyway we can get away with not using boost libraries here?

pxr/imaging/plugin/hdRpr/distantLight.cpp
```diff
@@ -61,7 +61,6 @@ void HdRprDistantLight::Sync(HdSceneDelegate* sceneDelegate,
                 *dirtyBits = HdLight::Clean;
                 return;
             }
-            rprRenderParam->AddLight();
```
bsavery, line 64: I thought this was for dome light?  Or is it removing some default distant light?

hshakula: That was about [tracking the number of lights](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/renderParam.h#L53-L55) on the scene so that we can decide whether we need default light or not. It has been moved to HdRprApi class

pxr/imaging/plugin/hdRpr/light.cpp
hshakula: This exact block can be replaced with a wall of if-else:
```
if (m_light.which() == kLightTypePoint) {
    // point light logic
} else if (...) {
    ...
}
```

But it does not make sense. If we want to get rid of Boost then we need to remove it everywhere. Do you have any information about Boost in USD in the feature? Right now it's one of the mandatory requirements for building

PR #294, opened 2020-06-05
Expose shadow and reflection catcher controls

hshakula:


bsavery:
So how would the user mark an object as shadow catcher?  Is there some UI in Houdini?

hshakula:
So as these two attributes are attributes of materials there is no way to add RPR specific UI on the existing UI for UsdPreviewSurface and Principled shader. Right now to use it the user will have to use the `Edit Properties` node. I believe that we will create our specific material nodes in the future, or/and Houdini will implement runtime discover of renderer-specific attributes via custom schemas, will see.

Issue #210, opened 2019-12-19
Artifacts on refractive materials when opacity enabled


hshakula:
Without opacity:
![glass_ok](https://user-images.githubusercontent.com/22181845/71187304-7fe70f80-2287-11ea-88fa-f91d85603805.PNG)
With opacity:
![glass_wrong](https://user-images.githubusercontent.com/22181845/71187305-7fe70f80-2287-11ea-985c-44ad088cdd80.PNG)

PR #295, opened 2020-06-05
Improve AOV system

hshakula:
* Allow the user to bind any AOV supported by RPR
* Fix primId AOV: use `HdRprim::GetPrimId` as object id - effectively enables usdview's selection logic

    primId-selection

* Workaround RPR's rule that sounds like "every AOV is multisampled" by resolving singlesampled AOV (any ID AOV, worldCoordinate, etc) only once. Related Issue [FIR-1681](https://amdrender.atlassian.net/browse/FIR-1681).

    To be clear, by `singlesampled AOV` I mean AOV for which accumulation does not make any sense. The most outstanding example is `RPR_AOV_OBJECT_ID` - if a pixel on the edge of a few objects RPR will average IDs, this averaging will give you invalid ID.
    Here are [AOVs](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/c7e153d7eed8e8b035562beb97bcd991aba0872a/pxr/imaging/plugin/hdRpr/aovDescriptor.cpp#L54-L64) that I identified as singlesampled. Please verify.
* Allow hydra user to get raw unmodified color AOV.

    In hdRpr, we actually have two different kinds of AOVs - raw and computed. Raw AOV is an AOV that is fed to Hydra directly without any math applied to values (except maybe converting formats, e.g. from RPR's native HdFormatFloat32Vec4 to HdFormatFloat32Vec3 (e.g. RPR_AOV_UV)). Computed AOV is an AOV that requires some sort of computations, with raw AOVs as its inputs, before giving results to Hydra. Before the current changes, we had only one computed AOV - depth, which is actually the clip-space depth (aka OpenGL depth). Previously, there was no way for the user to simultaneously get denoised\tonemapped\alpha-composited color AOV and raw color AOV. These changes will allow this.

* Replace `HDRPR_DISABLE_ALPHA` environment setting with corresponding render setting to allow runtime changes.

    enable-alpha

You might ask why do we need to be able to disable alpha at all? Here are at least two points:
* You have more than one shadow/reflection catcher on the scene and you want to view them. You have two options: disable them all one by one or disable alpha.
* Opacity AOV is not bugfree in RPR, e.g. https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/issues/210 but it's already fixed. Here is another example of a still existing issue.

Alpha off
image_2020-06-05_16-09-43

Alpha on
image_2020-06-05_16-09-33

Opacity is clearly broken where the same ray simultaneously intersects volume shape and invisible sphere light (I want to point out that light's shape does not intersect volume's shape). Related issue [FIR-1682](https://amdrender.atlassian.net/browse/FIR-1682).

### DIFF
--- a/pxr/imaging/plugin/hdRpr/CMakeLists.txt
+++ b/pxr/imaging/plugin/hdRpr/CMakeLists.txt
@@ -139,6 +139,7 @@ pxr_plugin(hdRpr
         ${OptIncludeDir}
 
     PRIVATE_CLASSES
+        aovDescriptor
         rendererPlugin
         renderDelegate
         renderPass
--- a/pxr/imaging/plugin/hdRpr/aovDescriptor.cpp
+++ b/pxr/imaging/plugin/hdRpr/aovDescriptor.cpp
@@ -1,0 +1,144 @@
+/************************************************************************
+Copyright 2020 Advanced Micro Devices, Inc
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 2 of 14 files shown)


bsavery:
This looks great in general.

PR #296, opened 2020-06-09
Query UV primvar name from bound material

hshakula:
When mesh has few UV sets or it has non-standard (`st`) UV primvar name we must query the name of the needed UV primvar from bound material


bsavery:
Merge should have both token names?

PR #297, opened 2020-06-09
Add northstar support

hshakula:


hshakula:
Blocked by https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/295 - we need to disable opacity AOV in Northstar because it's always zeroed

PR #298, opened 2020-06-09
Add UDIM support

hshakula:


bsavery:
This depends on Northstar support no?

hshakula:
Yes, it works only with northstar plugin but this code can be compiled and used with any plugin, it will simply still miss images when UDIM is used

PR #301, opened 2020-07-01
Rework material handling

hshakula:
- Implement nodal data-driven material generation from the given HdMaterialNetwork.
- Prepare the ground for MaterialX support:
  * parse node definitions from the given .mtlx files
  * define all RPR native nodes (RPR_MATERIAL_NODE_*) as .mtlx files
- Implement convenient wrapper around arithmetic nodes to simplify its usage: previously we had to keep two distinct code-paths for scalar/vector inputs and texture inputs.
- Houdini: add VOP node for each RPR material node

The structure of the project has been changed a bit - hdRpr was split into two libraries: Hydra plugin library and common library used in both Hydra plugin and the Houdini VOP node plugin.


cmake/defaults/Packages.cmake
```diff
@@ -99,6 +99,8 @@ else()
     message(STATUS "Skipping OpenVDB support")
 endif()
 
+find_package(MaterialX REQUIRED)
```
hshakula, line 102: @bsavery, should we include MaterialX as submodule or expect the user to provide it, what do you think?

hshakula: I'm going to make it a submodule because on Linux we need to make sure that `_glibcxx_use_cxx11_abi` is set to 0 when building Houdini plugin.

pxr/imaging/rprUsd/materialNodes/mtlxFiles/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+set(MTLX_FILES
```
bsavery, line 1: could we do Patterns/*.mtlx?  Or do you want to list explicedly.

CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ project(hdRpr)
 
 cmake_minimum_required(VERSION 3.12)
 
+# Shut up 'Up-to-date' install messages.
+set(CMAKE_INSTALL_MESSAGE LAZY)
```
bsavery, line 6: Hmmm, this seems like a user choice no?

pxr/imaging/rprUsd/materialNodes/mtlxFiles/CMakeLists.txt
hshakula: Globbing in cmake is a bad smell. One of the reasons for it is that when you add a file cmake will not update the globbing query. You need to do it explicitly by resetting the cache. So listing your files explicitly, in general, is the better approach

CMakeLists.txt
hshakula: You are right, I will remove it

PR #302, opened 2020-07-06
Speed up basis curves creation

hshakula:
I was profiling RPR on the scene with a quite huge curve primitive and noticed that most of the basis curves creation time was spent in std::vector::reserve function.

A straightforward solution for this issue is to precalculate the amount of required memory before converting Hydra data to RPR data.

Before these changes, basisCurves creation time was 9057.5ms, after - 107.0ms.


bsavery:
I like 10x improvements.

PR #307, opened 2020-07-29
Parallelize texture loading 

hshakula:
# Texture processing

Texture loading and creation of `rpr::Image` is one of the most time-consuming tasks. This process consists of two key steps:
* Load image data from a disk
* Create `rpr::Image` from the loaded data

# Parallelize what can be parallelized 

The obvious idea here is to parallelize what can be parallelized.
`rpr::Image` creation cannot be parallelized due to the single-threaded nature of RPR API.
So the only thing left is loading from a disk.

## Performance measurements before and after parallelization

I'm testing on the [NVIDIA's USD Attic](https://developer.nvidia.com/usd) scene and the [SideFX's Bar](https://www.sidefx.com/contentlibrary/bar-scene/) scene. The following time values are measured from the delegate creation to the beginning of the first render call (end of SyncAll) - effectively time to the first pixel minus time of the first `rprContextRender`.

|  Tahoe           | Attic, sec  | Bar, sec |
| ---------------- | ----------- | -------- |
| Single-threaded  | 26.620      | 14.353   |
| Multi-threaded   | 8.358       | 9.543    |

|  Northstar       | Attic, sec  | Bar, sec |
| ---------------- | ----------- | -------- |
| Single-threaded  | 26.289      | 13.117   |
| Multi-threaded   | 8.558       | 8.197    |

As you can see, the Attic scene has a huge improvement in time in the same time the Bar scene not.
More specific measurements should explain why.

### Attic - 345 .png textures in total
  - Single-threaded loading of textures from a disk: 20.452 sec
  - Multi-threaded loading of textures from a disk: 02.068 sec
  - Northstar `rpr::Image` creation (always single-threaded): 03.478 sec
  - Tahoe `rpr::Image` creation (always single-threaded): 03.977 sec

### Bar scene - 3044 .exr textures in total
  - Single-threaded loading of textures from a disk: 03.711 sec
  - Multi-threaded loading of textures from a disk: 02.270 sec
  - Northstar `rpr::Image` creation (always single-threaded): 02.072 sec
  - Tahoe `rpr::Image` creation (always single-threaded): 07.809 sec

Despite 10 times the difference in the number of textures, .exr wins drastically but it does not parallelize so well as .png loading.
So these numbers answer why we see such a huge boost on the Attic scene when parallelizing texture loading and also they show why numbers on the Bar scene not that remarkable. But these numbers also bring at least two new questions:
* Why .png loading is so much slower?
* Why .exr loading has much lower parallelization potential?

The actual code that loads images from the disk is inside of USD, or rather inside of Glf module.

# PNG loading

PNG loading is done with stb_image library.

To estimate the performance price of GlfImage abstraction and its concrete implementation that uses stb_image I've made a simple test ([stbtest.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/4995706/stbtest.zip)) that uses stb_image directly and measured its performance. It showed me that the single-threaded performance of .png image loading through GlfImage abstraction is almost the same as raw stb_image loading. That means that if we want to improve .png images loading speed all we can do is improve decoder implementation.

It's possible to force GlfImage to use OIIO library to load .png images. OIIO library uses libpng for .png loading. This gives such results for the Attic scene:
  - Single-threaded loading of textures from a disk: 21.916 sec OIIO vs 20.452 sec stb_image
  - Multi-threaded loading of textures from a disk: 02.688 sec OIIO vs 02.068 sec stb_image

Short googling gave me this [note](https://caedesnotes.wordpress.com/2015/03/23/comparing-performance-stb_image-vs-libjpeg-turbo-libpng-and-lodepng/) in which the author compares stb_image, lodepng, libpng and libjpeg decoding performance. It says that "libpng is fastest, optimized stb_image takes about 33-40% longer" - this conclusion contradicts with my comparison stb_image vs OIIO (libpng). But the author of this note used libpng directly, probably this can explain such a result?

# EXR loading

EXR loading is done with OIIO  library (which uses openexr under the hood).

I did not find any public comparisons of the currently available solutions for loading .exr images. So I will have to do it myself. But I think that the current performance of .exr images is out of the question especially when comparing to .png loading performance.

# RAT loading

This is a native Houdini image format promoted as the best image format for texture mapping. Previously support for this format was implemented inside of hdRpr under ifdef that was enabled when compiling hdRpr as Houdini plugin. For the sake of image loading cleanliness and unification of Houdini and usdview plugin, I've moved .rat related code to the separate plugin that will be built only when Houdini's library is available.

# Conclusion

Regardless performance of the image decoders embedded in USD, parallelization of GlfImage loading gives us tangible results - ~3.2x less time to the first pixel on the Attic scene and ~1.5x speed up for the Bar scene.


bsavery:
Thank you for the VERY thorough description!

bsavery:
Do we want to address these points?

Why .png loading is so much slower?
Why .exr loading has much lower parallelization potential?

Moving from OpenImageIO to direct libraries?  I'm not sure if this is worth it.

hshakula:
I don't think it worth it either. Even if we will manage to speed up decoding at least 20%, which is unlikely, it will be insignificant in comparison to `rpr::Image` creation time as it's already taking more time now than decoding

PR #308, opened 2020-07-29
Add northstar specific rendering loop

hshakula:
Account for high dependency of Northstar performance on RPR_CONTEXT_ITERATIONS: higher value gives better performance.


bsavery:
Do we want this to be draft waiting for new SDK?

hshakula:
There is no reason for that. This PR compiles with the current SDK

PR #310, opened 2020-07-30
Add MaterialX material node

hshakula:
Only one input: .mtlx file path. And all it does is pass the content of the input file to `rprLoadMaterialX`.

image

In draft until the next RPR SDK release.


hshakula:
Ready for review, can be merged after https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/346

PR #314, opened 2020-08-03
Update to RIF SDK 1.5.4

hshakula:


pxr/imaging/plugin/hdRpr/rifcpp/rifContext.cpp
```diff
@@ -268,6 +268,7 @@ ContextMetal::ContextMetal(rpr::Context* rprContext, std::string const& modelPat
 }
 
 std::unique_ptr<Image> ContextMetal::CreateImage(HdRprApiFramebuffer* rprFrameBuffer) {
+#ifdef __APPLE__
```
hshakula, line 271: They reverted new behavior to the old one - https://github.com/Radeon-Pro/RadeonProImageProcessing/pull/392, so that `rifContextCreateImageFromMetalMemory` is not present in RadeonImageFilters lib.

PR #318, opened 2020-08-17
Refactor render qualities handling (+ enable Northstar on macOS)

hshakula:


pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
```diff
@@ -32,18 +32,18 @@ def hidewhen_not_ambient_occlusion_mode(render_setting_categories):
...
                 'values': [
-                    "Low",
-                    "Medium",
-                    "High",
-                    "Full"
+                    'Low',
+                    'Medium',
+                    'High',
+                    'Full',
+                    'Northstar'
```
bsavery, line 46: Rename to "full 2.0 (Beta)"

hshakula: Done

PR #319, opened 2020-08-18
Fix depth of field

hshakula:
Focal length and sensor size should be in millimeters


hshakula:
Before 
before

After
after

PR #342, opened 2020-09-02
Enable building hdRpr against py3 version of Houdini

hshakula:


pxr/imaging/rprUsd/contextHelpers.cpp
```diff
@@ -114,6 +114,7 @@ const std::map<RprUsdPluginType, const char*> kPluginLibNames = {
     {kPluginNorthstar, "Northstar64.dll"},
     {kPluginHybrid, "Hybrid.dll"},
 #elif defined __linux__
+    {kPluginNorthstar, "libNorthstar64.so"},
```
hshakula, line 117: Yeah, it does not belong here but, while testing Houdini-py3 changes on Linux, I noticed Northstar binary for Ubuntu.

So I activate northstar for Linux here instead of creating separate PR

PR #359, opened 2020-10-04
Fix batch rendering

hshakula:
* Fix rendering of singlesampled AOVs (e.g. depth)
* Remove `progressive` render setting and disable progressive render by default when in batch
    Here disabled progressive rendering means that we try to maximize the number of samples rendered per one `rprContextRender` call, e.g. setting `RPR_CONTEXT_ITERATIONS` to `m_maxSamples` when possible. The only benefit we had with progressive rendering in the batch session is that it was giving us the ability to update the current progress that is logged by husk/usdrecord. Now, as we are moving to RPR 2.0, we have `RENDER_UPDATE_CALLBACK` that allows us to maximize `RPR_CONTEXT_ITERATIONS` without losing updates on progress. Though, progress reporting is lost for Tahoe but: a) it should be really easy for the core team to implement `RENDER_UPDATE_CALLBACK` support for Tahoe (`FIR-1701`) and b) Tahoe is going to be a legacy option soon

Here are some measurements:
| Commit        | Kitchen  | Market   | Attic    |
| ------------- |:--------:|:--------:|:--------:|
| develop (63978ee)       | 00:50.80 | 01:44.22 | 10:20.83 |
| PR (adf9793)      | 00:14.85 | 00:45.58 | 04:42.29 |
|               |          |          |          |
| **Improvement** | x3.42 | x2.29 | x2.2 |

### DIFF
--- a/pxr/imaging/plugin/hdRpr/renderBuffer.cpp
+++ b/pxr/imaging/plugin/hdRpr/renderBuffer.cpp
@@ -12,6 +12,7 @@ limitations under the License.
 ************************************************************************/
 
 #include "renderBuffer.h"
+#include "renderDelegate.h"
 #include "renderParam.h"
 #include "rprApi.h"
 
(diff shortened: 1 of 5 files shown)


bsavery:
Needs a merge fix.  Also with the removal of progressive updates, can we still get mid-render images saved from husk?  RenderMan has a similar feature:  https://rmanwiki.pixar.com/display/REN/Checkpointing+and+Recovery and it is quite useful in production.

hshakula:
Recovery as-is is not possible to support with the current RPR API.

As far as I know, checkpointing is possible with husk, they call it snapshotting. I'll give it a try and come back here

bsavery:
Needs a merge fix.

PR #379, opened 2020-10-27
Disable SetAOVindexLookup when tracing is enabled

hshakula:
It pollutes and blows up tracing play files. This leads to reduced readability.
So as SetAOVindexLookup makes sense only on the plugin side we can safely disable it when tracing is enabled.


bsavery:
Seems like a minor issue but fine.

hshakula:
Atsushi’s request

PR #382, opened 2020-10-28
Integrate RPRMtlxLoader

hshakula:


bsavery:
No description here.  This is or loading a mtlx file directly?  Or mtlx network passed through hydra?

hshakula:
By default, hydra does not know anything about materialX. At least the hydra from the latest USD dev.
That's why materialX support is implemented through a custom rpr node - `rpr_materialx_node`. In USD it looks as following:
```
def Material "material"
{
    token outputs:rpr:surface.connect = </material/shader.outputs:surface>

    def Shader "shader"
    {
        uniform token info:id = "rpr_materialx_node"
        asset inputs:file = @./material.mtlx@
        token outputs:surface
    }
}
```

PR #386, opened 2020-10-29
Fix batch rendering

hshakula:
This PR is functionally identical to https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/359.

The only difference is the code. In https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/359 we have one `RenderImpl+RenderUpdateCallback` function pair that simultaneously handles both code paths for an interactive session (houdini/usdview viewports) and a non-interactive session (husk, usdrecord). This PR splits them into two different pairs for much easier reading (much less branching and code are specific to session type). This PR seems like a better version of the same functionality as it's much easier to support and change if needed because you don't need to think about both session modes at once.

@bsavery, let me know which one do you prefer.

And now a bit more about the content of these pull requests.
As you might remember, you have expressed [doubt](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/359#issuecomment-704398461) about getting intermediate results. And it's definitely not groundless. I have changed my initial implementation in such a way that it is now possible to use husk's snapshotting functionality. I really love how it works now.

Snapshotting is supported for Northstar differently from other rendering plugins. For Northstar, I rely on RENDER_UPDATE_CALLBACK. This allows me to maximize RPR_CONTEXT_ITERATIONS from the very beginning of the rendering process. For other rendering plugins, I just leave RPR_CONTEXT_ITERATIONS at one because there is no other way to keep track of the progress and allow taking snapshots.

The snapshotting itself implemented in the most obvious way: when the host application maps HdRenderBuffer for reading I block the current thread, notify the rendering thread that I want to resolve framebuffers, and just wait until it's done. There is no harm from blocking the mapping thread, at least for command-line rendering utilities such as husk and usdrecord.

Here are some recent time measurements:

Northstar
Old default - 01:29.32
New default - 46.67 (x1.91 improvement)
New without progress logging - 46.51
New with 2 snapshots - 47.49

Tahoe
Old default - 34.29
New default - 34.09 (within the margin of error)
New without progress logging - 33.87
New with 2 snapshots - 34.57

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 build/*
+pxr/imaging/plugin/hdRpr/python/__pycache__/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,8 +42,52 @@ if(NOT MaterialX_FOUND)
     set(MATERIALX_BUILD_GEN_GLSL OFF)
     set(MATERIALX_BUILD_GEN_OSL OFF)
     set(MATERIALX_BUILD_TESTS OFF)
-    add_subdirectory(deps/MaterialX)
+    add_subdirectory(deps/MaterialX EXCLUDE_FROM_ALL) # EXCLUDE_FROM_ALL allows us to skip installation of mtlx static libraries
+
+    install(
+        FILES
+            deps/MaterialX/libraries/bxdf/standard_surface.mtlx
+            deps/MaterialX/libraries/bxdf/usd_preview_surface.mtlx
+        DESTINATION
+            libraries/bxdf)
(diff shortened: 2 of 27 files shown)


bsavery:
I don't have strong feelings, but agree with this statement:  "This PR seems like a better version of the same functionality as it's much easier to support and change if needed because you don't need to think about both session modes at once."

hshakula:
Okay, let's stop on this variant then

PR #391, opened 2020-11-05
Handle mtlx material from Autodesk's Scene delegate

hshakula:
Autodesk's Hydra Scene delegate may give us a mtlx file path directly


deps/rprMtlxLoader/rprMtlxLoader.cpp
```diff
@@ -813,6 +813,16 @@ Node::Ptr Node::Create(mx::Node* mtlxNode, LoaderContext* context) {
     } else if (mtlxNode->getCategory() == "normal") {
         rprMaterialSystemCreateNode(context->rprMatSys, RPR_MATERIAL_NODE_INPUT_LOOKUP, &rprNode);
         rprMaterialNodeSetInputUByKey(rprNode, RPR_MATERIAL_INPUT_VALUE, RPR_MATERIAL_NODE_LOOKUP_N);
+    } else if (mtlxNode->getCategory() == "sqrt") {
```
bsavery, line 816: is this supposed to be part of change?

hshakula: adsk materials use sqrt extensively. So I've added support for this node through pow node.

These changes already in RPRAPI repo (https://github.com/Radeon-Pro/RadeonProRenderAPI/pull/95/commits/e58a70f010643061b1de245eea770c9763b8cf87) but I decided to add them here too so we can use them asap

PR #396, opened 2020-11-12
Expose contour rendering mode

hshakula:
### PURPOSE
Support contour rendering with Northstar.

### EFFECT OF CHANGE
Added new render mode - **Contour**, and its render settings (linewidth of a particular edge detection feature and others). Added new geometry setting - **Ignore Contour**, that controls whether contour should be extracted for a mesh or not.

* Global Illumination mode
image

* Contour mode
image

### TECHNICAL STEPS
* Refactored code generation scripts to extend its functionality:
    * More generic way to define dynamic UI menus.
         * It allows us to dynamically add Contour render mode entry depending on the current render quality:
         image image

    * Add collapsible folders.
         * It allows us to group contour settings into one collapsible folder
            image image

* Improved HdRprApiAov creation. Now we set the DirtyAOVRegistry flag when HdRprApiAov is released. Previously it was not required because HdRprApiAov could have been released only when we set the DirtyAOVRegistry flag because of a new bound AOVs. Now AOVs can be created/released when the contour render mode is enabled/disabled (see UpdateRenderMode in rprApi.cpp).

### DIFF
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -360,6 +360,7 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
         }
     }
 
+    bool isIgnoreContourDirty = false;
     bool isVisibilityMaskDirty = false;
     if (*dirtyBits & HdChangeTracker::DirtyPrimvar) {
         HdRprGeometrySettings geomSettings = {};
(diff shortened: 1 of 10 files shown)


hshakula:
In a draft, until the next RPR SDK release because it will allow us to enable/disable contour render mode dynamically (with the current RPR SDK we need to resync the whole scene if we want to enable it).

PR #397, opened 2020-11-13
Fix motion blur exposure

hshakula:
### PURPOSE
To fix the motion blur difference between RPR and Karma.

* RPR
northstar

* Karma
karma

### EFFECT OF CHANGE
Correct motion blur camera exposure.

* RPR fixed
northstar_fixed

### TECHNICAL STEPS
Hardcode rpr_camera exposure to 1. Hydra always sample transforms in such a way that starting transform matches the camera's `shutterOpen` and ending transform matches the camera's `shutterClose`.


bsavery:
Bit odd that hydra normalizes the camera shutter that way.  Interested to know the design decision why, but I guess it makes sense.

hshakula:
I just noticed that such behavior was introduced only in USD 20.11. Do we want to keep old behavior for 20.08 and older?

PR #406, opened 2020-11-24
Improve ID AOVs

hshakula:
### PURPOSE
Previously, we were not able to distinguish 0 id pixels and void pixels (where no geometry present).
By offsetting the AOV index lookup by one we can achieve that void pixels will have -1 id.

### EFFECT OF CHANGE
Empty pixels now have a valid value (-1). It allows distinguishing empty pixels and pixels that contain id==0. For usdview, it enables deselection by pressing on empty space and avoids selection of the first object when selecting empty space.

### TECHNICAL STEPS
Do offset each AOV index lookup by one, thus allowing us to get id 1 for the first element and 0 for empty space. By offsetting it back, we get 0 id for the first element and -1 for empty space.


hshakula:
Before
image_2020-11-24_17-09-38

After
image_2020-11-24_17-11-29

PR #411, opened 2020-11-27
Add a way to override mesh ID

hshakula:
### PURPOSE
To allow the user to control the object ID that is rendered on the corresponding AOV.

### EFFECT OF CHANGE
Added `ID` parameter to the `Render Geometry Settings` LOP node.

### TECHNICAL STEPS
Parse id from the constant primvars and if it's authored override `HdRprim::GetPrimId`.


bsavery:
Not sure if this was supposed to be merged first but please fix conflicts.

PR #414, opened 2020-12-01
Improve .rpr export

hshakula:
### PURPOSE
Speed up .rpr export and reduce output size.

### EFFECT OF CHANGE
* Export .rpr sequence within a single process
  * Create RPR context only one time
  * Incrementally update scene instead of resyncing from scratch on each frame
* Export .rpr with RPRLOADSTORE_EXPORTFLAG_EXTERNALFILES flag by default
  * Allows reusing data within frames
  * Add new checkbox "Export As Single File" to revert to the old behavior

### TECHNICAL STEPS
* Added non-UI render setting `rprExportAsSingleFile` that controls usage of `RPRLOADSTORE_EXPORTFLAG_EXTERNALFILES`
* Added time-dependent render setting creation in LOP_RPRExportHelper for correct single-process export
* Update `RPR Export .rpr` HDA to reflect all changes


bsavery:
Seem the texture cache setting could be added here?

bsavery:
Ah lets do a separate PR for the SDK

PR #417, opened 2020-12-03
Update RPR to 2.1.8

hshakula:
### PURPOSE
To update RPR SDK to the most recent version and expose new functionality.

### EFFECT OF CHANGE
* Expose new AOVs:
    * Camera space shading normal
    * Cryptomatte AOVs (TODO integration)
* Allow exporting .rpr with compiled images instead of raw ones:
    Exporting compiled images greatly reduces Rprs render times but it can't be used if RprsRender and the current core do not match due to possible incompatibilities in the format of compiled images.


hshakula:
Right now, the checkbox for using image cache on .rpr export is on by default. With the latest info from Atsushi, I'm now leaning more toward disabling it by default. What do you think, @bsavery?

bsavery:
Yeah, I think it's tough to enable by default with all the caveats

PR #422, opened 2020-12-06
improve performance for multicore and multithread (SMT) CPUs

sidietz:
### PURPOSE
Performance optimization for zen2+ and bulldozer arch.

### EFFECT OF CHANGE
nothing.

### TECHNICAL STEPS
nothing.

### NOTES FOR REVIEWERS
Someone may display a warning that this optimization does not work on Haswell, IvyBridge and Intel (R) archs before GerminiLake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,17 +19,21 @@ Getting and Building the Code
         - Microsoft Visual Studio
     - CMake
     - Python
+    
+- Optional:
+    - git
 
 #### 2. Download the hdRpr source code
 
 You can use ```git``` to clone the repository.
+For modern, SMT or multi-threaded systems use:
 
 ```
 > git clone --recurse-submodules -j64 https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD
 Cloning into 'RadeonProRenderUSD'...
 ```
 
-or
+For legacy or low-power devices use:
 
 ```
 > git clone https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Getting and Building the Code
 You can use ```git``` to clone the repository.
 
 ```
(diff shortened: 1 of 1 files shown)


hshakula:
We have only 4 submodules. Could you please elaborate on why do we need to force everyone to setup 64 workers to just fetch 4 submodules?

git docs:
> -j <n>
--jobs <n>
    The number of submodules fetched at the same time. Defaults to the submodule.fetchJobs option.

Anyone who would customize the number of jobs already knows what he/she does and these numbers don't make sense to them.

@bsavery, thoughts?

bsavery:
Yeah, I agree with what @hshakula said.  -j4 is probably enough  Also I'm not sure how heavy IO tasks like git submodule always are the most efficient with high worker numbers but I could be wrong.

@sidietz Appreciate the contribution!  The changes here seems like it's not Zen specific, maybe change the titled?

sidietz:
Lets assume a non-real high-throughput SMT server with the following configuration:
64 cores, SMT2 (meaning 2 threads per core) resulting in 128 Threads in an 2HE 19 inch Rack with NVMe SSDs as primary disks.

That system should be able to handle -j 64. -j - as far as I know - does not only parallelize submodules. I think in the most recent git version it parallelizes the download streams, too.

However for legacy systems (less than 4 cores with 8 threads) another switch (like -j 8) should be considered valid.

bsavery:
> I think in the most recent git version it parallelizes the download streams, too.

Did you find any documentation of this?  I haven't seen that.  Interesting.

d-croft:
> > I think in the most recent git version it parallelizes the download streams, too.
> 
> Did you find any documentation of this? I haven't seen that. Interesting.

After having checked the doc you seem to be right.

bsavery:
Ah I realized this was for master after I merged it @d-croft in the future we do PR's to develop and then i merge to master when we bump the version.  So just do yours to develop please!

PR #429, opened 2020-12-22
Fix RPR trace generation

hshakula:
### PURPOSE
Recent changes (https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/379) lead to undesired behavior - RPR tracing had this line inside of it:
```
rprContextSetParameterByKey1u(context,RPR_CONTEXT_TRACING_ENABLED,(rpr_uint)1);  RPRTRACE_CHECK
```
It means that when you run the trace, the trace itself enables tracing.

So the purpose is to avoid leaking RPR_CONTEXT_TRACING_ENABLED into the trace.

### EFFECT OF CHANGE
None

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -2915,8 +2915,7 @@ private:
 
         m_isRenderUpdateCallbackEnabled = false;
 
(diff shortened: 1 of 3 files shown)


d-croft (approved):
Clean Code et. al. would approve, so do I.

PR #430, opened 2020-12-22
Fix rendering when color AOV is not bound

hshakula:
Since #413, color AOV is not created in advance.
From the other side, RPR refuses to render if color AOV is not created.
It's possible for Hydra to ignore color AOV.
For example, in usdview when depth AOV is selected, only depth AOV is bound.
That's why we must always create color AOV.

### PURPOSE
Fix errors on the call of `rprContextRender` when color AOV is not bound.

### EFFECT OF CHANGE
Fixed rendering when non-color AOV is enabled in usdview 20.08+.

### TECHNICAL STEPS
Create color AOV when initing the RPR context.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -276,6 +276,7 @@ public:
         try {
             InitRpr();
             InitRif();
+            InitAovs();
 
             {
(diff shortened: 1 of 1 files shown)


d-croft (approved):
Seems good to me. Initializing an object in it's constructor often helps dealing with runtime bugs.

PR #431, opened 2020-12-23
Improve MaterialX loader

hshakula:
### PURPOSE
Improve MaterialX loader

### EFFECT OF CHANGE
* Added support of MaterialX distance and angle units
* Added support of MaterialX non-LUT colorspaces

### NOTES FOR REVIEWERS
* Implementation for colorspace conversions taken from MaterialX stdlib glsl code.


d-croft (approved):

d-croft (commented):
Seems good to me on the first sight, but I am no CGI expert.

PR #432, opened 2021-01-04
Fix crash on save of .hip

hshakula:
### PURPOSE
To fix crash on save of .hip.

### EFFECT OF CHANGE
Fixed crash on save of .hip.

### TECHNICAL STEPS
Validate callback data before use.


d-croft (changes requested):
The check seems nice, but I'd recommend using not -1.
Is there a proper MACRO or static value like in the Linux kernel?
EINVAL, e.g.

hshakula:
> The check seems nice, but I'd recommend using not -1.
> Is there a proper MACRO or static value like in the Linux kernel?
> EINVAL, e.g.

Unfortunately, there is no documentation on `VOP_Node::opChanged` except [this](https://www.sidefx.com/docs/hdk18.5/_o_p___value_8h_source.html) (on line 52) and [this](https://www.sidefx.com/docs/hdk18.5/class_v_o_p___node.html#ab12c7fbf127c3f46e151c47fdd32c6ae). So there are no constants (or definitions) in HDK that would more expressively describe `-1`.

Issue #433, opened 2021-01-10
Displacement not working with RPR


matiascodesal:
Here's a simple repro that I tested in usdview.  It's a plane with a uniform 2 unit displacement.  If I render it in usdview with hdStorm, the plane is offset 2 units above the origin.  When I switch to RPR, the plane drops down to the origin.

**Repro:**
[displacement_repro.zip](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/files/5792163/displacement_repro.zip)
1. Exctract displacement_repro.zip
1. Make sure grid_scene_repro.usda and grid.usdc are in the same directory.
1. usdview grid_scene_repro.usda

hshakula:
https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/434 fixes this issue. But your exact scene would crash with the Full quality (RPR 2.0) because the material does not define texture coordinate primvar and hence quad mesh created without them. In the current RPR SDK, such setup (mesh with displacement without texture coordinates) leads to a crash. It's already fixed internally, waiting for the next RPR SDK release.

btw, if you need displacement asap my advice would be to bake displacement and use  UsdUvTexture as displacement input.

matiascodesal:
Thanks for the fix!

PR #434, opened 2021-01-11
Fix UsdPreviewSurface displacement defined with a constant

hshakula:
### PURPOSE
To fix https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/issues/433

### EFFECT OF CHANGE
Fixed UsdPreviewSurface displacement defined with a constant

### TECHNICAL STEPS
Use the output of the displacement node instead of the displacement node itself as a displacement output of the UsdPreviewSurface node.


d-croft (approved):
Refactoring shared to member access seems valid to me and should improve multithreaded performance.

PR #439, opened 2021-01-29
Update  RIF to 1.6.1

Vsevolod1983:
### PURPOSE
RIF is outdated

### EFFECT OF CHANGE
We have new RIF 1.6.1 integrated

### NOTES FOR REVIEWERS
We need to cast RIF_IMAGE_INTERPOLATION_NEAREST to (int) in order to avoid ambiguties. This macro became unsigned int (0u) in the new Rif and it might be treated as a null pointer as well.


cmake/modules/FindRif.cmake
```diff
@@ -15,7 +15,7 @@ endmacro(SET_RIF_VARIABLES)
 if(APPLE)
     SET_RIF_VARIABLES(OSX)
 elseif(WIN32)
-    SET_RIF_VARIABLES(Windows)
+    SET_RIF_VARIABLES(Windows/Dynamic)
```
hshakula, line 18: Add this inside of `SET_RIF_VARIABLES` to fix compilation on all platforms, not just on windows

hshakula (changes requested):

cmake/modules/FindRif.cmake
Vsevolod1983: Done

hshakula (approved):

PR #440, opened 2021-02-04
Add human-friendly colorized ID AOVs

hshakula:
### PURPOSE
Request from Japanese animation company. As far as I understood, they are using ID AOVs with the same style in other RPR plugins.

By default, ID AOV is of Integer type. Houdini's visualization of Integer images displays simply white and black.

Even if the ID AOV type set to float4 it would be rendered as a consecutive set of red-green colors. That's really hard to visually separate between two consecutive IDs.

materialId

This commit implements 3 new AOVs for each ID AOV: materialIdMask, objectIdMask, objectGroupIdMask. These AOVs are simply convenience layers over default ID AOVs. Each pixel of which calculated as a hash of the pixel's ID.

materialIdMask

### EFFECT OF CHANGE
3 new AOVs of float4 type: materialIdMask, objectIdMask, objectGroupIdMask. More human-friendly versions of default ID AOVs.

### TECHNICAL STEPS
Implement new class - `HdRprApiIdMaskAov ` - that wraps base ID AOV and calculates colorized ID AOV with the help of RIF user-defined filter.

### NOTES FOR REVIEWERS
In draft, until I add a corresponding unit test on CIS.


bsavery:
So quick question.  Do we really need "Human readable" aovs here?  I only really care that they work as intended in compositing packages like Nuke.  If this change fixes that then ok.

hshakula:
Well, I think that having these AOVs in our plugin is redundant. But Shin from the core team insisted that it would be very helpful for a few animation studios from Japan.

PR #467, opened 2021-04-21
Fixed cyrillic typos in some .mtlx files.

bnagirniak:
### PURPOSE
There are several typos with cyrillic characters in some .mtlx files. They were found during importing RPR MaterialX nodes into Blender USD plugin.

### EFFECT OF CHANGE
Fixed typos with cyrillic characters in some .mtlx files


hshakula:
Hehe, interesting, I was copying this text from the docs.

bnagirniak:
Then in the docs we have same typos.

PR #468, opened 2021-04-27
Update MaterialX to 1.38

hshakula:
### PURPOSE
To include the latest fixes to mtlx stdlib and other stuff in hdRpr.

### EFFECT OF CHANGE
MaterialX 1.37 is not supported anymore. See [changelist](https://www.materialx.org/assets/MaterialX.v1.38.Changelist.pdf).


hshakula:
Outdated. Both Hybrid and Northstar add their own loaders. `deps/rprMtlxLoader` should be deleted.

PR #471, opened 2021-05-03
Add support of deformation motion blur

hshakula:
### PURPOSE
Adds support of deformation motion blur

### EFFECT OF CHANGE
* Added support of deformation motion blur in the Full quality.
* New geometry render setting introduced - "Geometry Time Samples". The number of sub-frame samples to compute when rendering deformation motion blur over the shutter open time. The default is 1 (sample only at the start of the shutter time), giving no deformation blur by default. If you want rapidly deforming geometry to blur properly, you must increase this value to 2 or more. Note that this value is limited by the number of sub-samples available in the USD stage being rendered. Set to 0 to get all sub-samples available in the USD stage.

### TECHNICAL STEPS
* Use `HdSceneDelegate::SamplePrimvar` instead of `HdSceneDelegate::Get` to query the required amount of time samples from the scene delegate.
* Add new geometry setting - `rpr:geomSamples`.

### NOTES FOR REVIEWERS
Support for UsdSkel deformation blur can be quite easily added for USD 21.05+ only. Because there was no appropriate API to get few samples for a `HdComputation` before. It was only recently implemented and merged - https://github.com/PixarAnimationStudios/USD/pull/1455.

So I plan to open this pull request and move on to adding support of USD 21.05 in our plugin (there are many changes actually). After that, I will implement the UsdSkel deformation motion blur support.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -73,43 +73,6 @@ void HdRprMesh::_InitRepr(TfToken const& reprName,
     // No-op
 }
 
-template <typename T>
-bool HdRprMesh::GetPrimvarData(TfToken const& name,
-                               HdSceneDelegate* sceneDelegate,
-                               std::map<HdInterpolation, HdPrimvarDescriptorVector> const& primvarDescsPerInterpolation,
-                               VtArray<T>& out_data,
-                               VtIntArray& out_indices) {
-    out_data.clear();
-    out_indices.clear();
-
-    for (auto& primvarDescsEntry : primvarDescsPerInterpolation) {
-        for (auto& pv : primvarDescsEntry.second) {
-            if (pv.name == name) {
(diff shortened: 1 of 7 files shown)


bsavery:
Still in progress?

hshakula:
I don't plan to add anything to this PR. UsdSkel deformation blur will be in a separate PR.

bsavery:
My only concern is the motion samples setting.  People may intuitively set to 0 to get no motion blur.  However I don't see a better option here other than noting in the UI

hshakula:
Well, we can change samples range from [0; inf] to [1; inf] and remove this automatic behavior for 0 (when we get all available samples). At least because this does not make sense in the case of UsdSkel where we can generate an infinite amount of samples.

bsavery:
Ok I think that makes more sense.

bsavery:
Hey @hshakula can you fix the merge conflicts

hshakula:
@bsavery, done

PR #476, opened 2021-05-18
Made adjustments due to USD 21.05 changes:

bnagirniak:
### PURPOSE
After updating USD to 21.05 release there are following errors:
1. Light position: `Error in 'pxrInternal_v0_21__pxrReserved__::VtValue::_FailGet' at line 565 in file D:\amd-gpuopen\BlenderUSDHydraAddon\deps\USD\pxr\base\vt\value.cpp : 'Attempted to get value of type 'GfMatrix4d' from empty VtValue.'`
2. Camera: `[RPR ERROR] Failed to set camera focus distance -- invalid parameter in UpdateCamera at line 1809 of D:\amd-gpuopen\BlenderUSDHydraAddon\deps\HdRPR\pxr\imaging\plugin\hdRpr\rprApi.cpp(7adc67f)`

### EFFECT OF CHANGE
Fixed light positioning and log error with camera focus distance for USD 21.05.

### TECHNICAL STEPS
- fixed getting light transform
- added check for camera focus distance

### NOTES FOR REVIEWERS
Probably we need more checks here to support older USD versions. @hshakula ?

### DIFF
--- a/pxr/imaging/plugin/hdRpr/light.cpp
+++ b/pxr/imaging/plugin/hdRpr/light.cpp
@@ -397,7 +397,7 @@ void HdRprLight::Sync(HdSceneDelegate* sceneDelegate,
     HdDirtyBits bits = *dirtyBits;
 
     if (bits & DirtyBits::DirtyTransform) {
-        m_transform = GfMatrix4f(sceneDelegate->GetLightParamValue(id, HdTokens->transform).Get<GfMatrix4d>());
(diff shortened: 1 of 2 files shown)


hshakula:
> Probably we need more checks here to support older USD versions. @hshakula ?

The camera change is ok but the light one should be implemented with PXR_VERSION check

bsavery:
@hshakula approve?

hshakula:
@bsavery Yes

PR #491, opened 2021-07-19
Initial Baikal Next integration

hshakula:
### PURPOSE
To support compiling and using hdRpr with Baikal Next

### EFFECT OF CHANGE
Added Baikal next support 

### TECHNICAL STEPS
* A few tweaks to cmake scripts to support Baikal Next builds from cis (those builds have no rprloadstore library)
* Added new plugin type to the project

### NOTES FOR REVIEWERS
This is a very rough integration so that we could at least compile and use Baikal Next. The only test I did so far is rendered scene with one mesh. So don't expect all Baikal Next features to be present. Right now, Baikal Next is handled in the same way as Hybrid.

P.S. specifically for you @bnagirniak

### DIFF
--- a/cmake/modules/FindRpr.cmake
+++ b/cmake/modules/FindRpr.cmake
@@ -49,7 +49,7 @@ find_library(RPR_LOADSTORE_LIBRARY
     NO_SYSTEM_ENVIRONMENT_PATH
 )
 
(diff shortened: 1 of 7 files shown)


bnagirniak (approved):
As initial integration works ok

PR #493, opened 2021-07-22
Add support for aperture offset

### LINKED ISSUE #181: UsdGeomCamera's aperture offset is not working
I don't see right now how it should be implemented. All propositions are welcome

hshakula:
### PURPOSE
To improve the camera feature set. To resolve #181.

### EFFECT OF CHANGE
The camera's aperture offset is now supported.

### TECHNICAL STEPS
Copy implementation from the blender addon.

### NOTES FOR REVIEWERS
I've tested it in Houdini by comparing it to `Houdini GL` delegate renders.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -1949,13 +1949,20 @@ public:
         float focalLength;
 
         GfVec2f apertureSize;
+        GfVec2f apertureOffset(0.0f);
         HdRprCamera::Projection projection;
         if (m_hdCamera->GetFocalLength(&focalLength) &&
             m_hdCamera->GetApertureSize(&apertureSize) &&
+            m_hdCamera->GetApertureOffset(&apertureOffset) &&
             m_hdCamera->GetProjection(&projection)) {
             ApplyAspectRatioPolicy(m_viewportSize, aspectRatioPolicy.value, apertureSize);
             sensorWidth = apertureSize[0];
             sensorHeight = apertureSize[1];
+
+            apertureOffset[0] /= apertureSize[0];
+            apertureOffset[1] /= apertureSize[1];
+
+            apertureOffset[1] *= aspectRatio;
         } else {
             bool isOrthographic = round(m_cameraProjectionMatrix[3][3]) == 1.0;
             if (isOrthographic) {
@@ -1974,6 +1981,8 @@ public:
                 focalLength = m_cameraProjectionMatrix[1][1] / (2.0 * aspectRatio);
             }
         }
+
+        RPR_ERROR_CHECK(m_camera->SetLensShift(apertureOffset[0], apertureOffset[1]), "Failed to set camera lens shift");
 
         if (projection == HdRprCamera::Orthographic) {
             RPR_ERROR_CHECK(m_camera->SetMode(RPR_CAMERA_MODE_ORTHOGRAPHIC), "Failed to set camera mode");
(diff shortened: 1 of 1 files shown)


bsavery (approved):

pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1949,13 +1949,20 @@ class HdRprApiImpl {
...
+            m_hdCamera->GetApertureOffset(&apertureOffset) &&
             m_hdCamera->GetProjection(&projection)) {
             ApplyAspectRatioPolicy(m_viewportSize, aspectRatioPolicy.value, apertureSize);
             sensorWidth = apertureSize[0];
             sensorHeight = apertureSize[1];
+
+            apertureOffset[0] /= apertureSize[0];
+            apertureOffset[1] /= apertureSize[1];
+
+            apertureOffset[1] *= aspectRatio;
```
bnagirniak, line 1965: Multiplication is not need here

bnagirniak (changes requested):
Compared with HdStorm and there is mismatch in Y offset because of apertureOffset[1] multiplication.

pxr/imaging/plugin/hdRpr/rprApi.cpp
hshakula: Weird because to match with `Houdini GL` delegate this multiplication is needed. I will check Houdini's UI for other settings that might control this behavior

bnagirniak: yes, this is weird, because I tested in USD blender plugin and comparing to GL delegate this is not needed

hshakula: There is nothing in Houdini to control this behavior. What's more interesting `Karma` renderer results do not match with `Houdini GL`. So I don't know who is right here.

hshakula: `Cycles` renderer does not use aperture offset for a perspective camera at all. So I think the best would be to match with `Storm`.

bnagirniak (approved):

DagerD (approved):
Tested. Works ok.

pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1949,13 +1949,18 @@ class HdRprApiImpl {
         float focalLength;
 
         GfVec2f apertureSize;
+        GfVec2f apertureOffset(0.0f);
         HdRprCamera::Projection projection;
         if (m_hdCamera->GetFocalLength(&focalLength) &&
             m_hdCamera->GetApertureSize(&apertureSize) &&
+            m_hdCamera->GetApertureOffset(&apertureOffset) &&
```
bnagirniak, line 1956: Seems here apertureOffset isn't calculated for orthographic camera.

bnagirniak (changes requested):
Perspective camera works nice, it doesn't work for orthographic camera. 
apertureOffset == (0,0) for orthographic

pxr/imaging/plugin/hdRpr/rprApi.cpp
hshakula: I can confirm that apertureOffset is correctly passed in the case of the ortho camera. But it has no effect on the appearance of a scene. That's only the case for Northstar. Tahoe reacts to changes in lens shift.

Could you please try with the Full (Legacy) quality?

bnagirniak: Yes, I tried with Legacy quality - it works as expected, thank you for confirmation.

bnagirniak (approved):

pxr/imaging/plugin/hdRpr/rprApi.cpp
hshakula: https://amdrender.atlassian.net/browse/RPRNEXT-831

PR #496, opened 2021-07-24
Render device selection

hshakula:
### PURPOSE
To improve render device selection.

### EFFECT OF CHANGE
* Added a new GUI for render device selection. It can be found under the main menu in RPR->Render Devices.
image
* Removed the renderDevice render setting.

The device configuration is tied to the current hardware. By default, only one GPU is used, if available, otherwise all CPU threads are used. It's now possible to select any device combination. Render devices are configured per plugin (this is motivated by the fact that each particular plugin might behave differently for a different combination of hardware, e.g. Tahoe might be slower with CPU+GPU rendering while Northstar not, Hybrid does not support CPU rendering at all).
image

### TECHNICAL DETAILS
* Render device configuration GUI is implemented with PyQt.
* Render device configuration is saved under a separate cache folder as a JSON file.
* The same GUI code is reused between usdview and Houdini.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
+++ b/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py
@@ -215,25 +215,6 @@ render_setting_categories = [
         }
     },
     {
-        'name': 'Device',
-        'houdini': {
-            'hidewhen': hidewhen_hybrid
-        },
-        'settings': [
-            {
-                'name': 'renderDevice',
-                'ui_name': 'Render Device',
-                'help': 'Restart required.',
-                'defaultValue': 'GPU',
-                'values': [
-                    SettingValue('CPU'),
-                    SettingValue('GPU'),
-                    # SettingValue('CPU+GPU')
-                ]
-            }
-        ]
-    },
-    {
         'name': 'Denoise',
         'houdini': {
             'hidewhen': lambda settings: hidewhen_render_quality('<', 'High', settings)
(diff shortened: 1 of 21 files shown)


bsavery:
Would it make sense to have a CPU checkbox as well as the # of threads selector?

hshakula:
I was thinking about it but IMO less setting is better. Maybe you can elaborate on why the approach with two separate elements would make more sense here?

bsavery:
So usually users do not set number of CPU threads manually.  Having a checkbox allows them to not have to think about how many threads they have, or default "0" to all thread.  (Renderman allows -1 threads which is all but 1)

hshakula:
Okay, I will add a checkbox. Maybe there is anything else that you would like to tweak? So that I would do it in one sitting

hshakula:
image
image

When you enable the CPU number of threads is automatically set to the maximum value. The user can reduce it down to 1.

bsavery:
Looks perfect!

bsavery:
Actually, I just noticed something, why in the low - high settings does it cut off the "Super" in the description?  Something different with Vulkan?

hshakula:
These GPU names are taken from rpr_context by querying RPR_CONTEXT_GPUX_NAME. And Hybrid has a bug in rprContextGetInfo - it returns string size without null-terminating symbol. Tahoe and Northstar both return size including `\0`. I think it would be better to get it fixed on their side. I will make a PR.

hshakula:
https://github.com/Radeon-Pro/RPRHybrid/pull/890

PR #499, opened 2021-07-25
Fix compilation with Houdini 18.5.596+ on macOS

hshakula:
### PURPOSE
To fix compilation with Houdini 18.5.596+ on macOS

### EFFECT OF CHANGE
--

### TECHNICAL STEPS
Patch `HoudiniConfigVersion.cmake` that is shipped with the Houdini.


bsavery:
So just to be clear is this a bug in the Houdini cmake files?

hshakula:
The issue is coming from their recent changes in HoudiniConfig.cmake. By comparing the new config and the old one, I figured out that we can just roll back to the old behavior (without _houdini_use_framework) that works for us. Though I was not able to figure out why exactly their new approach does not work with hdRpr. So the only way to fix it was to patch the config.

PR #504, opened 2021-08-02
Improve adaptive sampling performance with RPR2 in batch mode

hshakula:
### PURPOSE
To improve adaptive sampling performance with RPR2 in batch mode

### EFFECT OF CHANGE
Improved performance of batch rendering when adaptive sampling enabled for the Full quality.

### TECHNICAL STEPS
Disable `RPR_CONTEXT_ACTIVE_PIXEL_COUNT` check for northstar.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -2446,9 +2446,10 @@ public:
         }
 
         // Though if adaptive sampling is enabled in a batch session we first render m_minSamples samples
-        // and after that render 1 sample at a time because we want to query the current amount of
+        // and after that, if needed, render 1 sample at a time because we want to check the current amount of
         // active pixels as often as possible
         const bool isAdaptiveSamplingEnabled = IsAdaptiveSamplingEnabled();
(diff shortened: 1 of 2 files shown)


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -2446,9 +2446,10 @@ class HdRprApiImpl {
         }
 
         // Though if adaptive sampling is enabled in a batch session we first render m_minSamples samples
-        // and after that render 1 sample at a time because we want to query the current amount of
+        // and after that, if needed, render 1 sample at a time because we want to check the current amount of
         // active pixels as often as possible
         const bool isAdaptiveSamplingEnabled = IsAdaptiveSamplingEnabled();
+        const bool isActivePixelCountCheckRequired = isAdaptiveSamplingEnabled && m_rprContextMetadata.pluginType == kPluginTahoe;
```
bsavery, line 2452: Adaptive sampling works for Northstar too.  Or do you just not check ever sample if Northstar here?

hshakula: We need to do a manual check only for Tahoe, that's what this flag is for.

PR #506, opened 2021-09-20
RPR 2.2.7  / Rif 1.7.1

Vsevolod1983:
### PURPOSE
Let's make some update regarding RPR and RIF


cmake/modules/FindRif.cmake
```diff
@@ -35,8 +35,11 @@ find_library(RIF_LIBRARY
 if(WIN32)
     set(RIF_BINARIES
         ${RIF_LOCATION_LIB}/dxcompiler.dll
+	${RIF_LOCATION_LIB}/dxil.dll
         ${RIF_LOCATION_LIB}/MIOpen.dll
         ${RIF_LOCATION_LIB}/RadeonImageFilters.dll
+        ${RIF_LOCATION_LIB}/OpenImageDenoise.dll
```
hshakula, line 41: Why do we need it?

hshakula (changes requested):

cmake/modules/FindRif.cmake
Vsevolod1983: I added everything RIF repo has

Vsevolod1983: At least RadeonML.dll is needed for sure. dxil.dll probably same

hshakula: Please remove this line, we don't use OIDN.

bsavery: What about for CPU + denoising?  Or at least we should discuss for future if that makes sense for final renders

hshakula: I only said that we don't use it at the moment so it makes no sense to add this 40 Mb dll to the package.

Surely we can use the OIDN if needed. It's more robust than the AI denoiser. Do you want me to add a way for users to select a denoiser filter and add support of the OIDN?

bsavery: What do we currently if CPU is enabled and denoising is turned on?  The RIF denoiser uses OIDN on cpu

hshakula: https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/rprApi.cpp#L2163

EAW denoiser is used in this case.

> The RIF denoiser uses OIDN on cpu

This is really surprising if it's true.

bsavery: Sorry @hshakula I should amend my comment "RIF AI Denoise" uses OIDN if run on CPU.

hshakula: What do you guys mean when you say RIF running on CPU? Do you able to create OpenCL context on CPU? For me, RIF never detected my CPU as a valid device.

bsavery: Let's keep it as is for now (no OIDN) and talk about this in a future task.

hshakula (approved):

PR #508, opened 2021-09-23
Add RPR USD schema

hshakula:
### PURPOSE
Add RPR USD schema

### EFFECT OF CHANGE
* All RPR specific attributes are exposed as a schema. See rprUsd API for more details.
* `RPR` environment variable is not needed anymore.

### TECHNICAL STEPS
* Added schema.usda file that defines all RPR specific staff: render settings, mesh properties, etc.
* Schema-generated files are generated at the build step.
* Removed the need in the `RPR` environment variable. Now we use rprUsd's plugin resource directory instead.

### NOTES FOR REVIEWERS
* Ideally, would be to query all of this ([1](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/python/generateRenderSettingFiles.py#L58-L637), [2](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/python/commonSettings.py#L13-L75), [3](https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/master/pxr/imaging/plugin/hdRpr/python/generateGeometrySettingFiles.py#L17-L62)) information from the schema. But this is not possible yet due to some limitations in the schema generation process. For example, properties order is lost after schema processing ([source](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/usdGeom/schema.usda#L116-L135) vs [generated](https://github.com/PixarAnimationStudios/USD/blob/release/pxr/usd/usdGeom/generatedSchema.usda#L19-L50))


pxr/imaging/rprUsd/schema.usda
```diff
@@ -0,0 +1,595 @@
...
+    customData = {
+        string className = "RendererSettingsAPI"
+        token apiSchemaType = "singleApply"
+        token[] apiSchemaAutoApplyTo = ["RenderSettings"]
+    }
+    doc = "Defines render settings for AMD Radeon Pro Renderer"
+)
+{
+    uniform token rpr:core:renderQuality = "Northstar" (
+        allowedTokens = ["Low", "Medium", "High", "HybridPro", "Full", "Northstar"]
```
bsavery, line 45: We need to come up with a better name than HybridPro for this

Full mode should be -> Northstar
Legacy -> Tahoe

bsavery:
Adding others to review as they may need to change plugins?

Vsevolod1983 (approved):

PR #512, opened 2021-11-08
USD schema fixes

hshakula:
### PURPOSE
To fix all issues introduced by https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/508 and other precedent changes.

### EFFECT OF CHANGE
* Fix CIS tests
* Add backward compatibility with Houdini 18.5
* Support USD 21.11

### NOTES FOR REVIEWERS
Splitting this into a few separate pull requests would be a wise choice. But it requires time which I don't have at the moment. We can discuss it if splitting is a must-have.


bsavery:
So this is still in draft mode.  What's the status?

s1lentssh:
> So this is still in draft mode. What's the status?

Currently fixing autotests to make all green, there were bugs in pipelines and assets

bnagirniak (approved):
Works with USD 21.11

bsavery:
Do we feel this is ready to merge?

s1lentssh:
We have problems with nested instancers, so for example Jet Engine scene becomes broken
I could take a look when finish Houdini PR's (probably today)

s1lentssh:
It's better to merge it now
We would fix instancer in separate PR

PR #514, opened 2021-12-09
Create parent dirs for shader cache folder

superkomar:
PURPOSE
Added method to create all necessary parent's directories.
The reason is CreateDirectory() throws an exception if any parent's folder does not exist

### DIFF
--- a/pxr/imaging/rprUsd/config.cpp
+++ b/pxr/imaging/rprUsd/config.cpp
@@ -4,6 +4,7 @@
 #include "pxr/base/tf/tf.h"
 #include "pxr/base/tf/instantiateSingleton.h"
 #include "pxr/base/tf/envSetting.h"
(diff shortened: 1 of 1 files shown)


hshakula:
What about using TfMakeDirs?

superkomar:
> What about using TfMakeDirs? https://github.com/PixarAnimationStudios/USD/blob/release/pxr/base/tf/fileUtils.h#L107

Ok, I'll change it.

hshakula:
Would not it be cleaner to remove archcreatedirectory at all and just use tfmakedirs?

superkomar:
> Would not it be cleaner to remove archcreatedirectory at all and just use tfmakedirs?

Done

hshakula (approved):
LGTM

PR #516, opened 2021-12-12
Simple gamma filter

s1lentssh:
### PURPOSE
Added simple gamma filter, in case if tonemap gamma is too complex

### EFFECT OF CHANGE
Added simple gamma filter

### TECHNICAL STEPS
- Added gamma into the schema and render settings generator
- Fixed hidewhen on some properties

### NOTES FOR REVIEWERS
Notes, if needed.


hshakula (approved):
LGTM

@bsavery, FYI, this filter is a request from Shin.

PR #518, opened 2021-12-13
Toon shader update

s1lentssh:
### PURPOSE
Updated toon shader

### EFFECT OF CHANGE
Added 5 colors mode and transparency logic

### TECHNICAL STEPS
Added new toon inputs from RPR SDK
Added runtime creation of transparent and blend nodes (require storing rpr context inside node)

### NOTES FOR REVIEWERS
Notes, if needed.

### DIFF
--- a/pxr/imaging/rprUsd/materialNodes/rpr/toonNode.cpp
+++ b/pxr/imaging/rprUsd/materialNodes/rpr/toonNode.cpp
@@ -24,14 +24,25 @@ TF_DEFINE_PRIVATE_TOKENS(_tokens,
     (color)
     (roughness)
     (normal)
+
     (shadowTint)
-    (midLevel)
-    (midLevelMix)
+    (midShadowTint)
     (midTint)
-    (highlightLevel)
-    (highlightLevelMix)
+    (midHighlightTint)
     (highlightTint)
+
+    (shadowLevel)
+    (midLevel)
+    (midHighlightLevel)
+    (highlightLevel)
+
+    (shadowLevelMix)
+    (midLevelMix)
+    (midHighlightLevelMix)
+    (highlightLevelMix)
+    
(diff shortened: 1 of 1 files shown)


pxr/imaging/rprUsd/materialNodes/rpr/toonNode.cpp
```diff
@@ -110,6 +148,28 @@ class RprUsd_RprToonNode : public RprUsd_MaterialNode {
             }
 
             return ProcessInput<GfVec3f>(id, value, m_toonClosureNode, RPR_MATERIAL_INPUT_NORMAL);
+        } else if (id == _tokens->colorsMode) {
+            if (value.IsHolding<int>()) {
+                int colorModeInt = value.UncheckedGet<int>();
+                return m_rampNode->SetInput(RPR_MATERIAL_INPUT_TOON_5_COLORS, colorModeInt) == RPR_SUCCESS;
+            }
+            TF_RUNTIME_ERROR("Input `%s` has invalid type: %s, expected - `Token`", id.GetText(), value.GetTypeName().c_str());
```
hshakula, line 156: expected token?

hshakula (approved):
I suppose that GUI is going to be quite dense with these changes - approximately 15 parameters. We will need to find a way to hide unneeded/unused parameters. This would require a different way of defining material nodes than we do right now and is definitely out of the scope of this PR.

pxr/imaging/rprUsd/materialNodes/rpr/toonNode.cpp
bsavery: ?

s1lentssh: Updated error messages

PR #519, opened 2021-12-13
Fix clipping range

DagerD:
### PURPOSE
UsdImagingLite doesn't send ClippingRange parameter to HdRPR, so we need to calculate it from projection matrix.

### EFFECT OF CHANGE
Add support of clipping range if UsdImagingLite is used.

### TECHNICAL STEPS
Added check for m_hdCamera->GetClippingRange


hshakula (approved):
LGTM in terms of parameter querying. Didn't check the math.

s1lentssh (approved):

PR #530, opened 2021-12-22
Fix building against monolithic USD

hshakula:
### PURPOSE
Fix building against monolithic USD

### EFFECT OF CHANGE
--


bnagirniak (approved):
Tested - works good

s1lentssh:
@bnagirniak are we sure it works? Locally for me it couldn't calculate USD version, so couldn't determine PXR_LIB_PREFIX, so IMPORTED_LOCATION_RELEASE isn't correct
I've tried to fix it inside my branch, could you please take a look? #573

bnagirniak:
This PR can be closed, fix is merged in #573

PR #532, opened 2021-12-23
Do not show prototype if instancer has no instances

hshakula:
### PURPOSE
Do not show prototype if an instancer has no instances

### EFFECT OF CHANGE
Prototype mesh is no more shown when an instancer has no instances.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -645,11 +645,6 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
                         }
                     }
                     m_rprMeshInstances.clear();
-
-                    auto visibilityMask = GetVisibilityMask();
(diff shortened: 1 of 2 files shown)


bsavery:
Is the prototype shown if there ARE instances?  Or is that already handled correctly?

hshakula:
> Is the prototype shown if there ARE instances? Or is that already handled correctly?

https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/42f331b96ffd9926c76e1a825799e130b3542aeb/pxr/imaging/plugin/hdRpr/mesh.cpp#L700-L701

Not shown. Though I’m not sure what is going to happen when instancer is removed. I’ll check.

hshakula:
> Though I’m not sure what is going to happen when instancer is removed. I’ll check.

Prototype mesh visibility was not updated and thus it stayed invisible. I fixed it in 270aacd by moving visibility handling in one place and handling all cases correctly.

PR #540, opened 2022-01-31
Fix build on windows

DagerD:
### PURPOSE
Fix build on windows broken after https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/538

### EFFECT OF CHANGE
Build on windows without errors.

### TECHNICAL STEPS
Turned out backslashes were needed for windows, added check for platform.


hshakula:
But #538 builds correctly, according to CIS. What error do you get exactly?

hshakula:
> But #538 builds correctly, according to CIS. What error do you get exactly?

It fixes the non-Houdini builds. That's why CIS builds are still valid and working.

hshakula (approved):

PR #541, opened 2022-02-01
Enabling Sphere light support for HybridPro

bnagirniak:
### PURPOSE
HybridPro supports rprContextCreateSphereLight().

### EFFECT OF CHANGE
Enabled HdRPR to craete Sphere light instead of box which is a workaround.

### TECHNICAL STEPS
Enabled IsSphereAndDiskLightSupported() for HybridPro either.


hshakula:
What about a disk light, does it work in HybridPro? If not, we should split the `IsSphereAndDiskLightSupported` into two separate functions.

bnagirniak:
I'll check this

bnagirniak:
> What about a disk light, does it work in HybridPro? If not, we should split the `IsSphereAndDiskLightSupported` into two separate functions.

Checked, disk light is also supported in HybridPro.

hshakula (approved):

DagerD (approved):
Tested. Works ok.

PR #543, opened 2022-02-16
Compilation fix gcc 9.3.0-17

s1lentssh:
### PURPOSE
Fix compile error on ubuntu

### EFFECT OF CHANGE
Updated c+= standard 11 => 14

### TECHNICAL STEPS
Patched CMake file

### NOTES FOR REVIEWERS
Notes, if needed.

### DIFF
--- a/cmake/defaults/gccclangshareddefaults.cmake
+++ b/cmake/defaults/gccclangshareddefaults.cmake
@@ -27,8 +27,8 @@
 # to remain minimal, marking the points where divergence is required.
 include(Options)
 
-# Turn on C++11; pxr won't build without it. 
-set(_PXR_GCC_CLANG_SHARED_CXX_FLAGS "${_PXR_GCC_CLANG_SHARED_CXX_FLAGS} -std=c++11")
+# Turn on C++14; pxr won't build without it. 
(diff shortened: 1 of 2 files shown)


s1lentssh:
> Any issues caused on other platforms or with blender / Maya plugin?

We'll check on monday, but it seems it's a valid fix, because we use std::make_unique in RadeonProRenderUSD/deps/rprMtlxLoader/rprMtlxLoader.cpp, which was intoduced in C++14

https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

It's interesting how it even compile right now

s1lentssh:
> > > Any issues caused on other platforms or with blender / Maya plugin?
> > 
> > 
> > We'll check on monday, but it seems it's a valid fix, because we use std::make_unique in RadeonProRenderUSD/deps/rprMtlxLoader/rprMtlxLoader.cpp, which was intoduced in C++14
> > https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
> > It's interesting how it even compile right now
> 
> https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/blob/develop/CMakeLists.txt#L30

Yes, but it still takes 11 standard in my configuration during clean build

s1lentssh:
Closed because duplicated in #561

PR #551, opened 2022-03-31
Make "alpha enable" switched off by default

Vsevolod1983:
### PURPOSE
We would like to see IBL visible in render by default. However, with alpha switched on by default we cannot see IBL background image what may confuse users.

### EFFECT OF CHANGE
IBL is visible in render by default.

### TECHNICAL STEPS
"Alpha Enabled" parameter affects visibility of IBL, so we make this parameter switched off by default.


Vsevolod1983:
Also @hshakula "hshakula" user is not available for some reason  to add him to the reviewers list

bnagirniak (approved):
Looks ok

s1lentssh (approved):

PR #555, opened 2022-04-25
Python launch fix for schema generation

Vsevolod1983:
### PURPOSE
For USDSchema generation we use hard coded command "python". However, not all python interpreters has "python" file as executable. For example in Maya it calls "mayapy.exe".

So, I suggest to use "python_executable" variable instead of hard coded value if set.

Feel free to suggest more handy solution

### EFFECT OF CHANGE
Changes made to project from the user's point of view.

### TECHNICAL STEPS
Changes made to code.

### NOTES FOR REVIEWERS
Notes, if needed.


Vsevolod1983:
@hshakula cannot assign you as a reviewer, please review

hshakula (approved):

PR #558, opened 2022-05-23
Improve houdini activator

hshakula:
### PURPOSE
Houdini activator does not work on some windows machines

### EFFECT OF CHANGE
Houdini activator now can find more potentially viable implicit HOUDINI_USER_PREF_DIR values.

### TECHNICAL STEPS
Check an additional viable location of implicit HOUDINI_USER_PREF_DIR


hshakula:
@bsavery @s1lentssh please review

bsavery:
I mean I wish there was a bit more clean way to do this than opening another process, but fine.

hshakula:
@bsavery, do you mean some sort of installer that would unzip and run this utility at the same time? It is def doable

PR #560, opened 2022-06-03
BLEN-111: Render time in viewport doesn't work properly

DagerD:
Fixed wrong setting parameters for render progress.


bnagirniak (approved):
Tested with USD Blender plugin - works good.

PR #561, opened 2022-06-04
 Added option to completely disable materialx

s1lentssh:
### PURPOSE
HybridPro has conflict with MaterialX library inside plugin, therefore it must be optional

### EFFECT OF CHANGE
MaterialX dependency now optional

### TECHNICAL STEPS
Added new parameter to CMake scripts

### NOTES FOR REVIEWERS
Notes, if needed.


andrey--p (approved):

cmake/modules/FindUSDMonolithic.cmake
```diff
@@ -75,23 +75,23 @@ if(USDMonolithic_FOUND)
         )
     endif()
 
-    add_library(usd_ms SHARED IMPORTED)
-    set_property(TARGET usd_ms APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
+    add_library(usd_usd_ms SHARED IMPORTED)
+    set_property(TARGET usd_usd_ms APPEND PROPERTY IMPORTED_CONFIGURATIONS RELEASE)
```
hshakula, line 79: Why do you need to rename the target? There are no reasons for that. If we really need to rename it, then something like "usd_monolithic" would be more expressive

pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -3339,6 +3339,9 @@ Don't show this message again?
         // TODO: verify
         if (RprUsdIsHybrid(m_rprContextMetadata.pluginType)) {
             RPR_ERROR_CHECK(m_rprContext->SetParameter(rpr::ContextInfo(RPR_CONTEXT_ENABLE_RELAXED_MATERIAL_CHECKS), 1u), "Failed to enable relaxed material checks");
+            RPR_ERROR_CHECK(m_rprContext->SetParameter(rpr::ContextInfo(RPR_CONTEXT_TONE_MAPPING), RPR_TONE_MAPPING_ACES), "Failed to set tonemapping");
+            RPR_ERROR_CHECK(m_rprContext->SetParameter(rpr::ContextInfo(RPR_CONTEXT_PT_DENOISER), RPR_DENOISER_SVGF), "Failed to set denoiser");
+            //RPR_ERROR_CHECK(m_rprContext->SetParameter(rpr::ContextInfo(RPR_CONTEXT_RESTIR_GI), RPR_TRUE), "Failed to set restir gi");
```
hshakula, line 3344: Do we need such a configuration for all users of the plugin or only in webusdviewer?

hshakula, line 3344: Do we need such a configuration for all users of the plugin or only in "the viewer"?

cmake/modules/FindUSDMonolithic.cmake
```diff
@@ -9,7 +9,7 @@ find_path(USD_INCLUDE_DIR pxr/pxr.h
     NO_SYSTEM_ENVIRONMENT_PATH)
 
 find_path(USD_LIBRARY_DIR
-    NAMES "${PXR_LIB_PREFIX}usd_ms${CMAKE_SHARED_LIBRARY_SUFFIX}"
+    NAMES "${PXR_LIB_PREFIX}usd_usd_ms${CMAKE_SHARED_LIBRARY_SUFFIX}"
```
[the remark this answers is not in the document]
s1lentssh: Done

hshakula (approved):

bsavery:
So just to understand this correctly, wouldn't this mean we have to do a separate build of USD for hybrid pro and Northstar?

s1lentssh:
> So just to understand this correctly, wouldn't this mean we have to do a separate build of USD for hybrid pro and Northstar?

Hm, most likely yes, because MaterialX in Hybrid Pro conflicts with MaterialX inside HdRpr.

s1lentssh:
Was fixed on Hybrid side, so don't need this feature anymore

PR #566, opened 2022-07-12
Open exr static option for maya

Vsevolod1983:
### PURPOSE
In Maya USD project we want to build hdRPR against built-in PixarUSD library. For this we need new option to customize build process in regards OpenEXR library. It got built as static lib in Maya's built-in PixarUSD library.


cmake/modules/FindOpenEXR.cmake
```diff
@@ -72,6 +72,10 @@ foreach(OPENEXR_LIB ${OpenEXR_FIND_COMPONENTS})
     find_library(OPENEXR_${OPENEXR_LIB}_LIBRARY
         NAMES
             ${OPENEXR_LIB}-${OPENEXR_MAJOR_VERSION}_${OPENEXR_MINOR_VERSION}
+
+if (MAYAUSD_OPENEXR_STATIC)
```
s1lentssh, line 76: Code highlighting broken here, so its very suspicious

Please make sure you could use "if" in "find" statements and make sure non-Maya builds aren't broken

hshakula: This `if-endif` closure inside a function call seems really weird and obscure. And I can actually explain why it works, ping me if you want to know it. In the meantime, I strongly recommend removing the if-endif and just leaving `${OPENEXR_LIB}-${OPENEXR_MAJOR_VERSION}_${OPENEXR_MINOR_VERSION}_s` as an additional option. Or even better to rewrite it this way:
```
if(MAYAUSD_OPENEXR_STATIC)
    set(OPENEXR_LIB_SUFFIX s)
endif()
find_library(...
NAMES
    ${OPENEXR_LIB}-${OPENEXR_MAJOR_VERSION}_${OPENEXR_MINOR_VERSION}${OPENEXR_LIB_SUFFIX}
    ${OPENEXR_LIB}${OPENEXR_LIB_SUFFIX}
...
```

Vsevolod1983: I see. I would prefer just to add "_s" as a new possible name in NAMES section because we use dynamic zlib library to be linked from static version of exr, so we need all these options in NAMES.

hshakula (approved):

s1lentssh (approved):

bnagirniak (approved):

PR #568, opened 2022-07-24
Update cmake to build with Houdini 19.5

hshakula:
### PURPOSE
To be able to compile hdRpr for Houdini 19.5

### EFFECT OF CHANGE
Houdini 19.5 support

### TECHNICAL STEPS
* Houdini 19.5 uses python 3.9 as the main python (boost bindings and other stuff). So we just need to look for this version too in `FindHoudiniUSD.cmake` file.
* Houdini 19.5 updated its OpenEXR libraries, so they are called differently now; and a few additional include files must be added too because of OpenEXR library internal rearrangement (as I can see they use more forward declarations now).
* rprUsd schema `Default` token has to be renamed to `default` because of a change in the usdGenSchema script - with `Default`, it generates tokens.h file with a token named `default` that is C++ reserved keyword which inevitably leads to compilation error.

### NOTES FOR REVIEWERS
This pull request is marked as a draft one because it depends on https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/557. This PR is required for Houdini 19.5 because it uses USD 22.05.


bsavery:
I merged #557 what now?

hshakula:
Now it's ready for review

PR #573, opened 2022-08-15
Added HybridPro for Linux

s1lentssh:
### PURPOSE
Previously we haven't had HybridPro in linux builds

### EFFECT OF CHANGE
Added HybridPro support for Linux

### TECHNICAL STEPS
- Added HybridPro.so entry for possible plugins map under Linux
- Fixed OpenEXR compile error
- Fixed FindUSDMonolithic (added version computation and fixed for USD >= 2111)

### NOTES FOR REVIEWERS
Notes, if needed.


cmake/defaults/Packages.cmake
```diff
@@ -110,6 +110,9 @@ if(HoudiniUSD_FOUND)
 else()
     # We are using python to generate source files
     find_package(PythonInterp 3.7)
+
+    # Required, otherwise couldn't find it
+    set(OPENEXR_LOCATION ${USD_INCLUDE_DIR}/../)
```
Vsevolod1983, line 115: Could you pass this variable into cmake  instead of hardcoding it ? We setup this variable in Maya via passing it to cmake script

Vsevolod1983 (approved):

bnagirniak (approved):
Looks good

PR #574, opened 2022-08-29
Scene units

V-Gorash:
### PURPOSE
HdRPR render delegate always interprete scene units as meters

### EFFECT OF CHANGE
Now HdRPR can accept "metersPerUnit" setting

### TECHNICAL STEPS
RprApi now has new field with scene unit scale matrix. It's being applied for all transforms


bnagirniak (approved):
Looks good

PR #575, opened 2022-08-30
Device check fix

dmitrii-z:
### PURPOSE
Skip unsupported GPUs to avoid later failures
Fix issues occurring in no-GPU systems


bsavery:
So in the non-GPU system does it just use CPU?

dmitrii-z:
> So in the non-GPU system does it just use CPU?

Yes, except Hybrid which do not support CPU

bnagirniak (approved):
Looks good

pxr/imaging/rprUsd/contextHelpers.cpp
```diff
@@ -429,12 +525,15 @@ RprUsdDevicesInfo RprUsdGetDevicesInfo(RprUsdPluginType pluginType) {
 
     RprUsdDevicesInfo ret = {};
 
-    if (RprUsdIsHybrid(pluginType)) {
+    if (pluginType == kPluginHybrid) {
```
s1lentssh, line 528: I think lets revert to RprUsdIsHybrid()

pxr/imaging/rprUsd/contextHelpers.cpp
```diff
@@ -62,6 +62,83 @@ const char* k_RadeonProRenderLibName = "libRadeonProRender64.dylib";
 const char* k_RadeonProRenderLibName = "libRadeonProRender64.so";
 #endif
 
+#ifdef HDRPR_ENABLE_VULKAN_INTEROP_SUPPORT
```
s1lentssh, line 65: Could you put this check out of HDRPR_ENABLE_VULKAN_INTEROP_SUPPORT
Basically let's always link vulkan and use it, probably we don't need HDRPR_ENABLE_VULKAN_INTEROP_SUPPORT option anymore

pxr/imaging/rprUsd/CMakeLists.txt
```diff
@@ -170,9 +172,9 @@ GroupSources(rprUsd)
 
 if(RPR_ENABLE_VULKAN_INTEROP_SUPPORT)
     target_link_libraries(rprUsd ${Vulkan_LIBRARIES})
```
s1lentssh, line 174: Probably linkage should always be outside RPR_ENABLE_VULKAN_INTEROP_SUPPORT, please check

pxr/imaging/rprUsd/contextHelpers.cpp
```diff
@@ -177,6 +177,23 @@ rpr_int GetPluginID(RprUsdPluginType pluginType) {
     return pluginID;
 }
 
+bool CheckCompatibility(const std::string& name) {
+    // the purpouse of this check is to avoid some "virtual" GPUs for witch contect could be created, but crashes when used
+    std::string lowercaseName = name;
+    std::transform(lowercaseName.begin(), lowercaseName.end(), lowercaseName.begin(), [](unsigned char c) { return std::tolower(c); });
+    
+    return (lowercaseName.find("amd") != std::string::npos ||
```
s1lentssh, line 185: Maybe make this check in loop? Like use initializer list of strings and make check on body

pxr/imaging/rprUsd/contextHelpers.cpp
```diff
@@ -429,12 +449,15 @@ RprUsdDevicesInfo RprUsdGetDevicesInfo(RprUsdPluginType pluginType) {
...
 
-    if (RprUsdIsHybrid(pluginType)) {
+    if (pluginType == kPluginHybrid) {
         ret.cpu.numThreads = 0;
 
-        std::string name = GetGpuName(pluginID, RPR_CREATION_FLAGS_ENABLE_GPU0, RPR_CONTEXT_GPU0_NAME, cachePath.c_str());
-        if (!name.empty()) {
-            ret.gpus.push_back({0, name});
+        HybridSupportCheck check;
+        if (check.supported(0)) {
```
s1lentssh, line 456: Should we check if any GPU supported, or only zero index?

s1lentssh (approved):

PR #576, opened 2022-08-30
Mtlx import fix

dmitrii-z:
### PURPOSE
fix for https://amdrender.atlassian.net/browse/RPRUS-106

### EFFECT OF CHANGE
Takes USD version from pxrConfig.cmake


DagerD (approved):
Works fine.

PR #578, opened 2022-09-04
Scene units token rename

V-Gorash:
### PURPOSE
Houdini USD use "stageMetersPerUnit" token name to provide unit size data. We need to use token with same name. https://www.sidefx.com/docs/hdk/_h_d_k__u_s_d_hydra.html#HDK_USDHydraCustomSettings

### EFFECT OF CHANGE
Now scene unit token is renamed to be same with Houdini. Also, token added to hdRPR parameters schema.


DagerD (approved):
Tested - works.

Vsevolod1983 (approved):

PR #582, opened 2022-09-21
Allow materialX to read texture coordinates that are not named "st"

trebconnell:
### PURPOSE
See description.

### EFFECT OF CHANGE
If the mesh doesn't have a primvar named "st" then it will find the next primvar with the texture coordinate [roll](https://graphics.pixar.com/usd/dev/api/_usd__page__datatypes.html#Usd_Roles).

### TECHNICAL STEPS
See 'Effect of Change'

### NOTES FOR REVIEWERS
The primary downside to this approach is if the mesh has multiple texcoords it will pick just the first one, which may not be what the materialX needs.  But since hdRPR currently doesn't support multiple texcoords I think this is acceptable.

### DIFF
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -367,9 +367,18 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
         auto rprMaterial = material->GetRprMaterialObject();
 
         auto uvPrimvarName = &rprMaterial->GetUvPrimvarName();
+        // This will currently be empty for MaterialX.
         if (uvPrimvarName->IsEmpty()) {
             static TfToken st("st", TfToken::Immortal);
             uvPrimvarName = &st;
+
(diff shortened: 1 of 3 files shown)


bsavery:
RPR actually supports 2 texture coordinates FYI.  I'm not sure hdRPR does anything to populate the second one though.

trebconnell:
Ah cool.  I'll tweak this to load the first two then.  

I'm getting test errors in the CI, but I can't access the Travis to see what they are.  Are these expected failures?  I wouldn't expect this change to break anything.

trebconnell:
Enabling second uv is a bit more involved as rprApi doesn't currently support it so I'll have to add some routing there.  I will probably do as a follow up PR.

bsavery:
> Enabling second uv is a bit more involved as the rprApi wrapper doesn't currently support it.
> Shouldn't be too hard to route through, but will be larger than this PR

Ok, I wasn't sure if the rprAPI supported the second or yet.  

Let me see about the CI failures.

s1lentssh (approved):

PR #588, opened 2022-10-12
MaterialX performance improvement

V-Gorash:
### PURPOSE
MaterialX sync process takes a lot of time when scene has mane MaterialX nodes

### EFFECT OF CHANGE
Now sync speed is increased on about 40%

### TECHNICAL STEPS
In previous version common libraries are being loaded for each node. In this PR library loading moved to common class. Also now rpr node can be created from matX string without a temp file

### DIFF
--- a/pxr/imaging/rprUsd/materialRegistry.h
+++ b/pxr/imaging/rprUsd/materialRegistry.h
@@ -20,6 +20,8 @@ limitations under the License.
 #include "pxr/base/arch/demangle.h"
 #include "pxr/base/tf/singleton.h"
 
+#include <MaterialXCore/Document.h>
+
 #include <RadeonProRender.hpp>
 
 class RPRMtlxLoader;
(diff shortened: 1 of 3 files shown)


bsavery (approved):

pxr/imaging/rprUsd/materialRegistry.h
```diff
@@ -145,7 +147,7 @@ class RprUsdMaterialRegistry {
     std::unique_ptr<RPRMtlxLoader> m_mtlxLoader;
 #endif
 
-    std::string m_materialXStdlibPath;
+    MaterialX::DocumentPtr m_stdLibraries;
```
s1lentssh, line 150: Can we please load libraries during RPRMtlxLoader construction? It made code easier

V-Gorash: This has been discussed with Alexandr. We decided not to move lib loading to constructur because this class is a singleton. It can be initialized before other resources and it can cause incorrect lib loading

PR #590, opened 2022-10-18
Reder devices change

V-Gorash:
### PURPOSE
Plugin has render device settings for outdated Tahoe engine, but doesn't have device option for HybridPro

### EFFECT OF CHANGE
1) Tahoe (legacy Full) engine has been removed from available renderers list
2) Tahoe has been removed from render device settings
3) HybridPro has been added to render device settings


V-Gorash:
Git conflict with develop. It would be easier to close it and create new PR

PR #594, opened 2022-11-02
instanceId aov

V-Gorash:
### PURPOSE
hdRPR does not support instanceId AOV

### EFFECT OF CHANGE
Now instanceId AOV support is added

### TECHNICAL STEPS
Now we use RPR_AOV_OBJECT_ID as instanceID and RPR_AOV_MESH_ID as primId

### DIFF
--- a/pxr/imaging/plugin/hdRpr/mesh.cpp
+++ b/pxr/imaging/plugin/hdRpr/mesh.cpp
@@ -758,13 +758,7 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
             for (auto& rprMesh : m_rprMeshes) {
                 rprApi->SetMeshId(rprMesh, id);
             }
-            auto instanceIndices = sceneDelegate->GetInstanceIndices(GetInstancerId(), GetId());
-            HdTimeSampleArray<VtValue, 2> ids;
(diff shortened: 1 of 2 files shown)


ARozghon:
working as expected

s1lentssh (approved):

Vsevolod1983 (approved):

pxr/imaging/plugin/hdRpr/mesh.cpp
```diff
@@ -759,9 +759,9 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
                 rprApi->SetMeshId(rprMesh, id);
             }
             for (auto& instances : m_rprMeshInstances) {
-                for (auto& rprMesh : instances) {
-                    rprApi->SetMeshId(rprMesh, id);
-                }
```
bsavery, line 764: Shouldn't we be setting object ID to the id val?

V-Gorash: SetMeshId incapsulates rprShapeSetObjectId method. We set prim ID as mesh object ID. And we set instanc ID as instance object ID. And RPR_AOV_MESH_ID shows base shape ID as ID of all it's instances. With RPR_AOV_OBJECT_ID all IDs are unique

PR #605, opened 2022-11-20
Fix nogpu issue

V-Gorash:
### WHY
Customers need to launch HdRPR on CPU-only instances

### WHAT
Added support of envireonment variable which prevents creation of RIF context and RPR GPU context

### HOW
If RPRUSD_CPU_ONLY variable set to 1 (True), RIF context won't be created. Also it prevents creating RPR context on GPU (ticket RPRUS-99).

### DIFF
--- a/pxr/imaging/plugin/hdRpr/rprApi.cpp
+++ b/pxr/imaging/plugin/hdRpr/rprApi.cpp
@@ -1913,43 +1913,40 @@ public:
             m_dirtyFlags |= ChangeTracker::DirtyScene;
         }
 
-        if (preferences.IsDirty(HdRprConfig::DirtyMotionBlur) || force) {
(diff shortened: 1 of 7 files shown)


pxr/imaging/plugin/hdRpr/rprApi.cpp
```diff
@@ -1913,40 +1913,43 @@ class HdRprApiImpl {
             m_dirtyFlags |= ChangeTracker::DirtyScene;
         }
 
-        if (m_rprContextMetadata.pluginType == kPluginNorthstar) {
```
V-Gorash, line 1916: I remove condition `m_rprContextMetadata.pluginType == kPluginNorthstar` because it does not needed after removing Tahoe

Vsevolod1983 (approved):

bsavery:
@trebconnell This is ok for you guys?

s1lentssh (approved):

PR #606, opened 2022-11-22
cmake viewer fix

dmitrii-z:
### PURPOSE
Update cmake file to fix plugin build for viewer

### EFFECT OF CHANGE
viewer build doesn't work correctly when NO_DEFAULT_PATH and NO_SYSTEM_ENVIRONMENT_PATH are used.
As they used for linux build, they can be defined as linux only.


ARozghon:
fixed

Vsevolod1983 (approved):

PR #610, opened 2022-11-29
Enable rendering of per face material for HybridPro engine

DMotorygin:
WHY
Currently scenes where per face material feature is used are not rendered correctly by HybridPro.

WHAT
Such scenes will be rendered correctly by HybridPro engine.

HOW
Set context property for HybridPro context when it is created.


ARozghon:
need to fix Hybrid pro with material X first

PR #613, opened 2022-12-08
Dome light transform fix

V-Gorash:
### WHY
HdRPR incorrectly process dome light rotation transforms

### WHAT
This PR can fix issue with dome light rotations

### HOW
There was wrong order of transform matrix multiplication. Now it's fixed


ARozghon:
working as expected

Vsevolod1983 (approved):

PR #615, opened 2022-12-15
profile output

dmitrii-z:
### PURPOSE
Added rendering time output to use in autotests


Vsevolod1983 (changes requested):
I would prefer to have such output on plugin side, not on hdRPR side.
We have RenderStats which can be filled, like we do it for another render statistics. And this RenderStats can be read from a plugin.

bsavery:
@Vsevolod1983  We do need profile for running standalone though,

Vsevolod1983 (approved):

Vsevolod1983:
hmm, ok, I see. But I would split business logic and log output somehow though

PR #617, opened 2022-12-20
Fix assigning materials per face

DMotorygin:
WHY
Currently scenes where per face material feature is used are not rendered correctly by HybridPro and crashes NorthStar.

WHAT
Such scenes should be rendered correctly by both render engines.
This bug happens because wrong path to material is used internally for per face materials when trying to find Sprim corresponding to such material.

HOW
Fix to the bug is getting correct (full) path of such material and using it to acquire material instead.


ARozghon:
approved by QA

pxr/imaging/plugin/hdRpr/mesh.cpp
```diff
@@ -652,7 +654,18 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
...
+						bool hasMaterialByShortPath
+							= (sceneDelegate->GetRenderIndex().GetSprim(HdPrimTypeTokens->material, m_geomSubsets[i].materialId)) != nullptr;
+
+						// geomSubset has only relative material path. Relative to mesh.
+						// materials however are stored by full material path
+						// so when relative material path is passed material is not found
+						// thus we have to get full material path to get pointer to material
+						SdfPath parentMesh = this->GetId().GetParentPath();
+						SdfPath relativeMaterialPath = m_geomSubsets[i].materialId.MakeRelativePath(SdfPath::AbsoluteRootPath());
+						SdfPath fullMaterialPath = parentMesh.AppendPath(relativeMaterialPath);
```
hshakula, line 666: Not used if `hasMaterialByShortPath==true`. SdfPath manipulations are not free and for some big scenes it can add up to a noticeable amount of time

pxr/imaging/plugin/hdRpr/mesh.cpp
```diff
@@ -361,6 +361,8 @@ void HdRprMesh::Sync(HdSceneDelegate* sceneDelegate,
     }
 
     // We are loading mesh UVs only when it has material
+	std::string dbgPath = m_materialId.GetAsString();
+
```
hshakula, line 365: Is it intended to stay in PR?

DMotorygin: debug code. removed

pxr/imaging/plugin/hdRpr/mesh.cpp
DMotorygin: done

hshakula (approved):

PR #625, opened 2023-01-18
Add version info to package

V-Gorash:
### WHY
Plugin package for Houdini does not contain version info

### WHAT
Now version info added to package metadata

### HOW
Information is being copied from version file during installation


ARozghon:
image
fixed

PR #627, opened 2023-01-23
Points material override

V-Gorash:
### WHY
In hdRPR Points objects does not support applying materials

### WHAT
Nowoverride of default material can be applied to points, also normals had been added to make sphere surfase more round

### DIFF
--- a/pxr/imaging/plugin/hdRpr/points.cpp
+++ b/pxr/imaging/plugin/hdRpr/points.cpp
@@ -153,7 +153,7 @@ void HdRprPoints::Sync(
             auto& topology = UsdImagingGetUnitSphereMeshTopology();
             auto& points = UsdImagingGetUnitSphereMeshPoints();
 
-            m_prototypeMesh = rprApi->CreateMesh(points, topology.GetFaceVertexIndices(), VtVec3fArray(), VtIntArray(), VtVec2fArray(), VtIntArray(), topology.GetFaceVertexCounts(), topology.GetOrientation());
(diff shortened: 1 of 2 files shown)


pxr/imaging/plugin/hdRpr/points.cpp
```diff
@@ -146,7 +153,7 @@ void HdRprPoints::Sync(
             auto& topology = UsdImagingGetUnitSphereMeshTopology();
             auto& points = UsdImagingGetUnitSphereMeshPoints();
 
-            m_prototypeMesh = rprApi->CreateMesh(points, topology.GetFaceVertexIndices(), VtVec3fArray(), VtIntArray(), VtVec2fArray(), VtIntArray(), topology.GetFaceVertexCounts(), topology.GetOrientation());
+            m_prototypeMesh = rprApi->CreateMesh(points, topology.GetFaceVertexIndices(), points, topology.GetFaceVertexIndices(), VtVec2fArray(), VtIntArray(), topology.GetFaceVertexCounts(), topology.GetOrientation());
```
V-Gorash, line 156: For sphere vertice and normal coordinates are same, so we can use points array as normals

pxr/imaging/plugin/hdRpr/points.cpp
```diff
@@ -23,7 +24,7 @@ limitations under the License.
 PXR_NAMESPACE_OPEN_SCOPE
 
 HdRprPoints::HdRprPoints(SdfPath const& id HDRPR_INSTANCER_ID_ARG_DECL)
-    : HdPoints(id HDRPR_INSTANCER_ID_ARG)
+    : HdRprBaseRprim(id HDRPR_INSTANCER_ID_ARG)
```
V-Gorash, line 27: We need to extend HdRprBaseRprim to get material id

ARozghon:
working as expected, approved

PR #628, opened 2023-01-31
Replace multiple calls to AOVIndicesLookup

V-Gorash:
### WHY
rprContextSetAOVindicesLookup function is faster than rprContextSetAOVindexLookup

### WHAT
Now rprContextSetAOVindexLookup is replaced with rprContextSetAOVindicesLookup. We can use single call to set up all index colors.


ARozghon:
approved by QA

dmitrii-z:
Reviewed

PR #630, opened 2023-01-31
shading normal aov removed from UI

dmitrii-z:
### PURPOSE
As Richard said, RPR_AOV_VIEW_SHADING_NORMAL  AOV is not supported in neither HybridPro nor Northstar, so I just removed it from UI until we have support.


ARozghon:
approved by QA

V-Gorash (approved):

PR #631, opened 2023-02-02
rat files conversion

dmitrii-z:
### PURPOSE
Resolves https://amdrender.atlassian.net/browse/RPRHOUD-41
As RPR does not support .rat files, this PR converts them to .exr and replaces the path passed to CreateEnvironmentLight()
The converted file is placed in the directory of the original .rat file, or, if it fails, in case of, for instance, write protection, into subdir in HDRPR_CACHE_PATH_OVERRIDE.


ARozghon:
approved by QA

V-Gorash (approved):

PR #637, opened 2023-02-15
Web material library

V-Gorash:
### WHAT
AMD online material library support had been added to Houdini plugin


ARozghon:
approved by QA, checked manually and all working as expected.
for now not support editing for WML materials. will be added with one of the future PR

dmitrii-z:
materialLibrary.py:124: Will it generate an absolute or relative path? If absolute, what will happen when the scene opens on another machine?
materialLibrary.py:124: I'm not sure that endless retrying is a good idea, maybe it's better to introduce an attempt counter?

dmitrii-z:
Approved

PR #638, opened 2023-02-16
Restore mtlx string loading

V-Gorash:
### WHY
In latest Hybrid Pro release loading materialx as string was fixed

### WHAT
This PR enables loading materialX as string without temp files


ARozghon:
approved by QA, looks good on Maya and hdRPR side

PR #642, opened 2023-02-28
Core 3.01.00

V-Gorash:
### WHY
Core v3.01.00 had been released

### WHAT
1) Update of core to 3.01.00
2) Add new precompiled core kernels
3) Add option to use legacy openCL backend


ARozghon:
Approved by QA

PR #652, opened 2023-03-31
IMath is required for bulding hdRPR against USD in maya 2024

Vsevolod1983:
### PURPOSE
IMath lib is required for hdRPR built against Maya 2024.
So we need just add it to cmake scripts


s1lentssh (approved):

bnagirniak (approved):
Looks ok

PR #655, opened 2023-04-07
sync and cache time

dmitrii-z:
### PURPOSE
Resolves: https://amdrender.atlassian.net/browse/RPRUS-158

Cache creation time is considered as the first render iteration time when the cache does not exist on renderer initialization. The iteration time is negligible in this case.
Sync time is considered as the time between the first call of sync and the renderer start.


ARozghon:
Approved by QA

PR #657, opened 2023-04-20
emission mode fix

dmitrii-z:
### PURPOSE
Resolves https://amdrender.atlassian.net/browse/RPRUS-164

### TECHNICAL STEPS
The emission mode property setting requires 1 for single and 2 for double-sided mode; the default behavior for comboboxes/enums is to use an integer for each option starting from 0. Thus, for some inputs, the value must be tuned.


ARozghon:
Approved by QA

PR #663, opened 2023-05-08
update Core 3.1.1 with performance update

Vsevolod1983:
### PURPOSE
RPR Core 3.1.1 update with performance fix for NS

### EFFECT OF CHANGE
Performance update for NS


ARozghon:
approved by QA

PR #674, opened 2023-07-03
Fix for Bg Color Override for plugins other than Houdini

Vsevolod1983:
### PURPOSE
Background color override fix for plugin other than Houdini


ARozghon:
approved by QA

bsavery (approved):

PR #682, opened 2023-08-05
standard_surface thin_walled param hack

dmitrii-z:
### PURPOSE
Resolves https://amdrender.atlassian.net/browse/RPRHOUD-70

### TECHNICAL STEPS
VtValue does not support ‘bool' type, so when the materialx document for the renderer is building, boolean parameters are represented as ints. In fact, there is only one boolean input in all supported mtlx nodes - thin_walled in standard_surface, so here is an exception for it.

### DIFF
--- a/pxr/imaging/rprUsd/hdMtlxFixed.cpp
+++ b/pxr/imaging/rprUsd/hdMtlxFixed.cpp
@@ -380,12 +380,19 @@ HdMtlxCreateMtlxDocumentFromHdNetwork_Fixed(
 
         const std::string & mxInputName = currParameter.first.GetString();
         mx::InputPtr mxInput = mxShaderNode->addInput(mxInputName);
-        
+
         // Convert the parameter to the appropriate MaterialX input format
         VtValue hdParamValue = currParameter.second;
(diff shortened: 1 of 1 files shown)


pxr/imaging/rprUsd/hdMtlxFixed.cpp
```diff
@@ -380,13 +380,20 @@ HdMtlxCreateMtlxDocumentFromHdNetwork_Fixed(
...
         mx::InputPtr mxInput = mxShaderNode->addInput(mxInputName);
-        
+
         // Convert the parameter to the appropriate MaterialX input format
         VtValue hdParamValue = currParameter.second;
         if (hdParamValue.IsHolding<bool>()) {
             bool value = hdParamValue.UncheckedGet<bool>();
             mxInput->setValue(value);
         }
+        // We have a trouble here: VtValue does not support �bool' type.
```
bsavery, line 390: Please fix non ascii characters in comment

bsavery (changes requested):
Fix non ascii characters

bsavery (approved):

PR #684, opened 2023-08-09
uber emission color intensity

dmitrii-z:
### PURPOSE
Resolves https://amdrender.atlassian.net/browse/RPRHOUD-66

### TECHNICAL STEPS
The intensity implementation is that the emission color should be multiplied by the intensity value (that's how it works in Blender plugin
To get this, BaseNode is extended with a class for the Uber node that adds an extra multiply node before the actual emission color node and redirects Uber's emission color and intensity inputs to this node.

### DIFF
--- a/pxr/imaging/rprUsd/materialMappings.h
+++ b/pxr/imaging/rprUsd/materialMappings.h
@@ -108,6 +108,7 @@ bool ToRpr(TfToken const& id, uint32_t* out, bool pedantic = true);
     (uber_sheen_weight) \
     (uber_emission_color) \
     (uber_emission_weight) \
+    (uber_emission_intensity) \
     (uber_emission_mode) \
     (uber_transparency) \
     (uber_sss_scatter_color) \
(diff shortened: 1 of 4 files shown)


ARozghon (approved):
Approved by QA

bsavery:
How is an intensity different than the weight.  Internally they should be the same inside Uber.

dmitrii-z:
> How is an intensity different than the weight. Internally they should be the same inside Uber.

I took the blender plugin as an example. There, the weight is used to assign MATERIAL_INPUT_UBER_EMISSION_WEIGHT, and intensity is multiplied by color to get the final color. So, weight shows how strong the emission is, and intensity how bright its color is. Should it behave another way?

ARozghon:
weight - it's for Emissive texture maps and intensity - for more emission power, we have the same option in Uber for the Blender Plugin

V-Gorash (approved):

bsavery (approved):

PR #687, opened 2023-08-14
Restart render thread automatically after manipulations with HdRprApi

mixern6:
### PURPOSE
Restart render thread automatically after manipulations with HdRprApi

### EFFECT OF CHANGE
RprApiSafeWrapper is created when RprApi is requested for editing. It stops the render thread and resumes it when the editing is finished. The change takes into account that RprApi can be requested simultaneously from multiple threads, and restarts the render thread only when the last instance of RprApiSafeWrapper is destroyed.
This change fixes the problem of the render thread not restarting if no changes are made.


Vsevolod1983 (approved):

ARozghon:
Approved for Houdini and hdRPR sides

PR #699, opened 2023-09-13
Fix issue with tracing directory

V-Gorash:
### PURPOSE
Some trace files are being placed in current working directory. not in specified by environment variable

### EFFECT OF CHANGE
Now all files are placed in specified tracing directory

### TECHNICAL STEPS
There was incorrect order of Core API calls. We should firstly specify tracing dir and after that enable tracing


ARozghon:
Approved by QA

bsavery (approved):

PR #704, opened 2023-10-05
removed rif denoising and upscaling

dmitrii-z:
### PURPOSE
Resolves: https://amdrender.atlassian.net/browse/RPRUS-175

### EFFECT OF CHANGE
Denoising and upscaling with RIF removed
UI updated to support Hybrid denoising and upscaling


ARozghon:
Approved by QA

MrKepzie:
@bsavery Was trying to get the USD render delegate a spin in Autograph and binding some settings; I saw that almost all denoisers are removed from the Northstar (no OIDN) whereas it still seems available in the Blender addon. Is there a reason for removing the denoisers ?

PR #724, opened 2024-04-05
Update Houdini integration for version 20 and elaborate README

mehmetoguzderin:
### PURPOSE

This PR fixes compile issues that occur during build with version 20 of the Houdini engine. Additionally, a small note about Python version with HFS environment variable was put to make it easier for people to adjust to the build environment.

### EFFECT OF CHANGE

Makes it possible to use the plugin with version 20.

### TECHNICAL STEPS

In domeLight.cpp, a compile-time check occurs with `HDK_API_VERSION >= 20000000` to choose the correct type. This pattern is also visible in VOP_RPRMaterial.(h|cpp), where HDK_API_VERSION helps to distinguish. But in case the plugin source does not care about version compatibility, I think this code can be made more concise by not checking for version.

### NOTES FOR REVIEWERS

The build was confirmed working with an Indie installation of Houdini version 20.

Please let me know if there's a documented minimum supported Houdini or if it is commit based (for example, latest commit works on 19, merge will be only 20+ etc.), and I can simplify accordingly if that's a viable path.

Thank you!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,9 @@ project(hdRpr)
 
 cmake_minimum_required(VERSION 3.15)
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/defaults
                       ${CMAKE_SOURCE_DIR}/cmake/modules
                       ${CMAKE_SOURCE_DIR}/cmake/macros)
@@ -26,8 +29,6 @@ add_definitions(${_PXR_CXX_DEFINITIONS})
 set(CMAKE_CXX_FLAGS "${_PXR_CXX_FLAGS} ${CMAKE_CXX_FLAGS}")
 
 include(Public)
-
-set(CMAKE_CXX_STANDARD 14)
 
 add_subdirectory(deps)
 add_subdirectory(pxr/imaging)
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Cloning into 'RadeonProRenderUSD'...
 Provide USD in one of two ways:
 
 * An installation of USD. Define pxr_DIR to point to it when running cmake, if required. You can download USD to build yourself from [GitHub](https://www.github.com/PixarAnimationStudios/USD). It should be built with OpenEXR libraries and usdGenSchema tool.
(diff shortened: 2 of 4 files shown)


ShinFujieda:
Could you check if it can be built and work also on Ubuntu 22.04?
I have tried it on my side, but it didn't work.

mehmetoguzderin:
@ShinFujieda Sure thing, I'll take a look at it this week. Thank you for the information!

mehmetoguzderin:
@ShinFujieda I've sent a commit to fix the build issues in Ubuntu 22.04! The issue was mostly due to the C++ Standard version not propagating well to the compiler and the new SDK requiring C++17 at the very least.

Could you please try again with the latest commit of this branch to see if the update resolves the issue? If not, I will be prompt to help.

Sorry that it took a while to reproduce due to certain licensing limitations in the container environment, but I managed to find a way. From here on, I can coordinate and help better.

Thank you very much!

ShinFujieda:
@mehmetoguzderin Thanks for the update! But when I tried on my side with Houdini 20.0.653, it seems the build succeeded and I could see RPR as a renderer just after running Houdini.
image

However, once running RPR on the viewport, it disappears. Have you seen this kind of issue?
If not, could you tell me how to build on your side?
image

mehmetoguzderin:
@ShinFujieda thank you! Is the screenshot from Windows or Ubuntu?

ShinFujieda:
It's on Ubuntu 22.04

mehmetoguzderin:
@ShinFujieda I will check again and update about the situation. Thank you!

mehmetoguzderin:
@ShinFujieda I was able to reproduce this issue and confirm that it is not happening on Windows but only on Ubuntu. Is there a similar problem with Ubuntu execution in version 19? Or is this a newly emerged problem? Since I don't have previous version's environment ready, this would be helpful to debug. Thank you!

ShinFujieda:
Ok, thanks for confirming. Let me try it with houdini 19.5

mehmetoguzderin:
@ShinFujieda That'd be appreciated, thank you!

ShinFujieda:
@mehmetoguzderin Hmm it seems it crashes with a segmentation fault when rendering a scene with RPR on Houdini 19.5.805. 
image

Though the issues are not totally the same, we can see the problem with both plugins for H19 and H20 built on **Ubuntu 22.04**.
Then, it might be better to resolve this issue as another PR if you can manage it. Thanks.

mehmetoguzderin:
@ShinFujieda Thank you very much for the confirmation! Then maybe this PR can be merged to enable Houdini 20 on Windows while I take a look at resolving Ubuntu issues in existing versions in its own investigation. Appreciated

takahiroharada (approved):

PR #733, opened 2024-08-27
Enable builds for Houdini 20.5

mehmetoguzderin:


ShinFujieda:
I noticed the environment light (dome light with texture) is not shown in the viewport.

Houdini VK:
image

RPR:
image

pxr/imaging/plugin/hdRpr/rprApiAov.cpp
```diff
@@ -592,7 +592,7 @@ HdRprApiDepthAov::HdRprApiDepthAov(
                             vec4 alpha = ReadPixelTyped(alphaImage, coord.x, coord.y);
                             vec4 color = ReadPixelTyped(inputImage, coord.x, coord.y);
                             if (alpha.x == 0) {
-                                color = make_vec4(1.f, 1.f, 1.f, 1.f);
+                                color = make_vec4(0.99f, 0.99f, 0.99f, 0.99f);
```
ShinFujieda, line 595: Could you put some comments here for the reason 0.99 is set?
For example, "depth = 1.f doesn't work to compose opacity with Houdini 20.5 or newer."

Also, if the value = 1.f doesn't work, should we also set "dstHi" parameter in `m_remapFilter to 0.99f?
https://github.com/GPUOpen-LibrariesAndSDKs/RadeonProRenderUSD/pull/733/files#diff-8a2cc46ce27bad8acdca816dbd2cac97944de1275242d0c8c890b9e741425f32R611

mehmetoguzderin: Makes sense, thank you very much, added the comment!

mehmetoguzderin:
@ShinFujieda -san,

Thank you for the review! If I set `dstHi` to 0.99f, all the grid and gizmos disappear. Is it better to keep as 1.0 or still update?

ShinFujieda:
I see. then I don't know how it's treated, but maybe the 1.0 value is reserved for such special components.
Then, let's keep it as 1.0 for now.

pxr/imaging/plugin/hdRpr/rprApiAov.cpp
ShinFujieda: Thanks

ShinFujieda (approved):

